PR #24, opened 2018-03-01
Extract some utils from multiple packages into one place

ritave:


ritave:
@nickreynolds FYI: I've extracted `ChaiConfig` into a separate package

PR #28, opened 2018-03-01
TCR whitelist address

nickreynolds:
- Adding ability for JEC to whitelist a newsroom (if newsroom not currently in any stage of application process)
- Giving JEC-whitelisted newsrooms a grace period during which they cannot be challenged
- Removing superuser from ACL and replacing with owner from OpenZeppelin's Ownable contract
- Unit tests


packages/contracts/contracts/ACL.sol
```diff
@@ -1,41 +1,31 @@
...
-  modifier requireSuperuser() {
-    require(isSuperuser(msg.sender));
+  modifier requireRole(string role) {
+    require(isOwner(msg.sender) || hasRole(msg.sender, role));
     _;
   }
 
-  modifier requireRole(string role) {
-    require(isSuperuser(msg.sender) || hasRole(msg.sender, role));
+  modifier requireOwner() {
```
ritave: `Ownable` already has `onlyOwner()`, use that

PR #30, opened 2018-03-02
add civil-editor

walfly:
still battling typescript, but the basics are there


walfly:
tsc is working now

packages/civil-editor/src/typings.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'slate';
+declare module 'slate-react';
+declare module '@storybook/addon-storyshots';
```
ritave: newline

ritave:
Also @walfly if there's stuff to be done concerning the package, or want track it or get some external help, go ahead and add issues to the tracker

PR #36, opened 2018-03-07
Added a Gnosis Multisignature Smart Contract

ritave:
Put it in a seperate package for the GPL license


packages/multisig/contracts/Migrations.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity ^0.4.19;
```
nickreynolds: remove ^

PR #45, opened 2018-03-16
Walkerflynn/ch72/civil editor should be able to create and

walfly:
This is still WIP.

left to do:
- all the tool buttons
- edit interfaces for some components
-  hotkeys
- schema and validation
- setting editable on root


dankins:
license should be Apache 2.0 in package.json. I reviewed what is there and lgtm. Will do an actual Review once it is ready to merge.

walfly:
Issues not resolved in this PR. 

- missing custom validation for drop cap and lead in
- toggle display styles for images

packages/editor/src/Editor.stories.tsx
```diff
@@ -2,37 +2,137 @@ import { storiesOf } from "@storybook/react";
...
 
-storiesOf("CivilEditor", module)
-    .add("main", () => {
-        const onChange = (value: any): any => { return; };
-        return (
-          <CivilEditor
+storiesOf("Civil Display View", module)
+    .add("paragraph", () => {
+        const initialValue = Value.fromJSON(paragraphJson);
+        return (<CivilEditor
```
dankins: if we are going to be adding more of these I think it makes sense to build a helper method to build this;
```
const makeEditor = (initialValue: any, readOnly: boolean) => (<CivilEditor
    value={initialValue}
    onChange={onChange}
    plugins={plugins}
    readOnly={readOnly}
  />);
```

packages/editor/src/plugins/components/Image.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { constants } from "../index";
+
+const ImageInline = styled.img`
+  width: 100%;
+  margin: 30px 0;
+`;
+
+const ImageBreakout = styled.img`
+  width: 90vw;
+  left: calc(-220px + calc(calc(805px - 90vw)/2));
```
dankins: today I learned something new - didn't know calc existed

packages/editor/src/plugins/components/ToolBar.tsx
```diff
@@ -0,0 +1,229 @@
...
+    stroke: #30E8BD;
+    fill: #30E8BD;
+  }
+`;
+
+const Button = styled.button`
+  background: transparent;
+  font-family: "Libre Franklin", sans-serif;
+  font-weight: 700;
+  color: #2B56FF;
```
dankins: so I think the answer to this is "we'll sort it out later", but there are a lot of hard coded colors in the styling. In SASS i would import a vars file, but not sure how it works in styled components. Is this something you plan to address in the future?

walfly: oh I had a note to myself to move them all to a constants file, but i forgot anyways

walfly: in the future we should probably import them from another package, so they can stay consistent across products

walfly:
I'll let @juliahimmel weigh in before I merge

walfly:
I published as 1.1

walfly:
adding my own block,

can't remove quotes

PR #46, opened 2018-03-19
Added newsroom factory

ritave:


packages/contracts/contracts/newsroom/NewsroomFactory.sol
```diff
@@ -0,0 +1,24 @@
...
+    multisigFactory = IMultiSigWalletFactory(_multisigFactory);
+  }
+
+  function create(address[] _owners, uint _required)
+    public
+    returns (Newsroom newsroom)
+  {
+    address wallet = multisigFactory.create(_owners, _required);
+    newsroom = new Newsroom();
+    newsroom.transferOwnership(wallet);
```
nickreynolds: we could just modify the newsroom contract to take an address in the constructor, I guess it doesn't matter too much though

ritave: `Ownable.sol` doesn't support taking address in constructor, might as well leave it explicit.

PR #52, opened 2018-03-23
Adding Header and Challenge functionality to debug-ui, modifying core to support

nickreynolds:
- creating header on debug ui pages that shows current Ethereum account and CVL balance of it
- Adding sections to newsroom page that render based on state of listing. Unhiding 'Appeal' section when listing is in Request Appeal Phase and render expiry time of phase.
- adding ability to request and withdraw voting rights in debug-ui (shown in header)
- adding support for appeal phases and challenge can be resolved in getListingState
- fixing bug in listing state function, properly rendering list of listings in challenge states
- updating event streams for commit vote and reveal vote phases to eliminate duplicate listings
- Adding ability to claim reward for challenges. On listing page, list out all IDs for challenges to this listing that have resolved.
- Adding "resolve post appeal" logic into "update status" transaction. No reason it needs to be its own. Listing status is updated from several different states.
- migrating TCR contracts


packages/core/src/artifacts/OwnedAddressTCRWithAppeals.json
```diff
@@ -1,1188 +1,12455 @@
...
...fffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200182815260200192505050602060405180830381600087803b15156128db57600080fd5b6102c65a03f115156128ec57600080fd5b50505060405180519050151561290157600080fd5b600160008086815260200190815260200160002060040160003373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060006101000a81548160ff021916908315150217905550833373ffffffffffffffffffffffffffffffffffffffff167ff01da32686223933d8a18a391060918c7f11a3648639edd87ae013e2e2731743836040518082815260200191505060405180910390a350505050565b6000600160008373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020600201549050919050565b6000600160008373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060010160009054906101000a900460ff169050919050565b600080600084815260200190815260200160002060040160008373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060009054906101000a900460ff16905092915050565b6000806000806000808781526020019081526
```
ritave: Wrong json uploaded. This one's for truffle debugging with lot's of unnecessary data. Use the stripped artifact

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -252,46 +353,174 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
+    const appealExpiryDate = new Date(appealExpiryUnixTimestamp.toNumber() * 1000);
+    const hasRequestedAppeal = await this.instance.getAppealRequested.callAsync(listingAddress);
+    return (!hasRequestedAppeal && appealExpiryDate > new Date());
+  }
+
+  /**
+   * Gets the expiry time of the request appeal phase.
+   * @param listingAddress Address of listing to check
+   */
+  public async getRequestAppealExpiryTimestamp(listingAddress: EthAddress): Promise<BigNumber> {
```
ritave: Do we ever need the timestamp, or should we just return dates for easier operating

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -252,46 +353,174 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
+  }
+
+  /**
+   * Gets the expiry time of the commit vote phase.
+   * @param listingAddress Address of listing to check
+   */
+  public async getCommitVoteExpiryTimestamp(listingAddress: EthAddress): Promise<BigNumber> {
+    // if there is no challenge
+    const challenge = await this.instance.getListingChallengeID.callAsync(listingAddress);
+    if (challenge.toNumber() !== 0) {
```
ritave:
```typescript
if (challenge.isZero()) {
  return challenge;
  // or even throw an error
}
const voting = ...
...
return revealPeriodActive;
```

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -252,46 +353,174 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
    */
   public async challengeExists(listingAddress: EthAddress): Promise<boolean> {
     return this.instance.challengeExists.callAsync(listingAddress);
   }
 
+  /**
+   * Gets the current challenge ID of the listing. Returns 0 if no current challenge.
+   * @param listingAddress Address of listing to check
+   */
+  public async getListingChallengeID(listingAddress: EthAddress): Promise<BigNumber> {
```
ritave: how about create a Challenge object/class and just operate on that, instead of having state managed in the TCR class

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -252,46 +353,174 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
+    return this.instance.getListingChallengeID.callAsync(listingAddress);
+  }
+
+  /**
+   * Gets whether or not the listing has an unresoved challenge
+   * @param listingAddress Address of listing to check
+   */
+  public async hasUnresolvedChallenge(listingAddress: EthAddress): Promise<boolean> {
+    const challengeID = await this.getListingChallengeID(listingAddress);
+    if (challengeID.toNumber() !== 0) {
```
ritave:
```typescript
if (challengeID.isZero()) {
  return false;
}
...
```

packages/debug-ui/scripts/tcrActions.ts
```diff
@@ -1,33 +1,138 @@
 import { Civil } from "@joincivil/core";
+import { getVoteSaltHash } from "@joincivil/utils";
 import BigNumber from "bignumber.js";
-import { EthAddress } from "../../core/build/src/types";
+import { Bytes32, EthAddress } from "../../core/build/src/types";
```
ritave:
```typescript
import { Bytes32, EthAddress } from "@joincivil/core";
```
that's why we're exporting the types in the first place

packages/debug-ui/scripts/tcrActions.ts
```diff
@@ -1,33 +1,138 @@
...
+export async function challenge(address: EthAddress, optionalCivil?: Civil): Promise<void> {
+  const civil = optionalCivil || new Civil();
+
+  console.log("Challenging TCR Listing");
+  const tcr = await civil.getDeployedOwnedAddressTCRWithAppeals();
+
+  const eip = await civil.getEIP20ForDeployedTCR();
+
+  const approvedTokensForSpender = await eip.getApprovedTokensForSpender(tcr.address);
+  if (approvedTokensForSpender.toNumber() < 1000) {
```
ritave:
```typescript
if (approvedTokensForSpender.isLessThan("1000"))
```

ritave: Also don't use constants here, since this script might work on a different parameters from parametizer

nickreynolds: The value 1000 is totally placeholder. Getting the actual value from parameterizer will be next thing I do (along with the rest of the parameterizer). I'll add a TODO here though.

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
nickreynolds: not really sure this is necessary right now. Would you think we should separate `Listing` into its own class as well?

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -63,6 +63,7 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
     return this.instance
       .NewListingWhitelistedStream({}, { fromBlock })
       .map((e) => e.args.listingAddress)
+      .distinct()
```
ritave: What if something gets delisted and listed again. Remove them leave `distinct()` to the library user in all cases as we can't know the use-case

ritave: Another example is listening on only latest events of any kind and refresh UI. Distinct would disable refreshing

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -208,19 +331,33 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
+      }
     }
     return isInApplicationPhase;
   }
 
+  /**
+   * Gets the application expiry timestamp for the given listing address
+   * @param listingAddress Address of listing to check
+   */
+  public async getApplicationExpiryTimestamp(listingAddress: EthAddress): Promise<BigNumber> {
```
ritave: Remove this and use Date as in other cases

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -229,21 +366,32 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
-    if (!commitPeriodActive) {
-      isInCommitPhase = false;
+    if (pollID) {
+      if (!challenge.equals(pollID)) {
+        isInCommitPhase = false;
+      }
+    }
+
+    if (isInCommitPhase) {
+      const voting = Voting.atUntrusted(this.web3Wrapper, await this.instance.voting.callAsync());
```
ritave: can we cache the addresses of related contracts?

ritave: also use `await this.getVotingAddress()` for less duplication

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -252,46 +400,172 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
+   * @param listingAddress Address of listing to check
+   * @throws {CivilErrors.NoChallenge}
+   */
+  public async getRevealVoteExpiryDate(listingAddress: EthAddress): Promise<Date> {
+    // if there is no challenge
+    const challenge = await this.instance.getListingChallengeID.callAsync(listingAddress);
+    if (challenge.isZero()) {
+      throw CivilErrors.NoChallenge;
+    }
+    const voting = Voting.atUntrusted(this.web3Wrapper, await this.instance.voting.callAsync());
```
ritave: use `this.getVotingAddress()`.

actually, might as well create `this.getVoting(): Promise<Voting>`

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
ritave: Add a clubhouse technical-debt card and let's merge this one as is. It would get too big

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -252,46 +400,172 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
+   * Challenge can be resolved if reveal period is over but request appeal phase is uninitialized
+   * @param address Address of listing to check
    */
   public async challengeCanBeResolved(listingAddress: EthAddress): Promise<boolean> {
-    return this.instance.challengeExists.callAsync(listingAddress);
+    const challengeID = await this.getListingChallengeID(listingAddress);
+    // if challenge exists, but not in commit or reveal phase, and request appeal phase not started
+    if (!challengeID.isZero()) {
+      const voting = Voting.atUntrusted(this.web3Wrapper, await this.instance.voting.callAsync());
+      const revealPeriodExpiryTimestamp = await voting.getRevealPeriodExpiry(challengeID);
```
ritave: Might as well use dates in Voting too

packages/core/src/utils/errors.ts
```diff
@@ -6,6 +6,7 @@ export enum CivilErrors {
   NoUnlockedAccount = "NO_UNLOCKED_ETHEREUM_ACCOUNT",
   NoPrivileges = "NOT_ENOUGH_PRIVILEGES_FOR_ACTION",
   EvmException = "EVM_EXCEPTION_OCCURED",
+  NoChallenge = "NO_CHALLENGE_FOUND_FOR_LISTING",
```
ritave: The goal for those errors is to have errors that can be caught in the UI and be handled and shown to the user with a nice explanation. I wonder if this one's too granular. How about `InvalidParameters` or similar?

packages/debug-ui/gulpfile.js
```diff
@@ -67,6 +67,7 @@ tsPipeline.registerBuildGulpTasks(
             index: __dirname + '/' + config.srcDir + 'code/indexEntry.ts',
             newsroom: __dirname + '/' + config.srcDir + 'code/newsroomEntry.ts',
             article: __dirname + '/' + config.srcDir + 'code/articleEntry.ts',
+            base: __dirname + '/' + config.srcDir + 'code/base.ts',
```
ritave: What's the difference between base and index?

ritave: Also, please don't use base.pug as an entry, it should be treated as an abstract class to be extended by other entries

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
nickreynolds: The point of this function is to get the set of currently whitelisted listings. If something is delisted and listed again it will be returned correctly. But this observable doesn't care which event is responsible for the whitelist. I intend to create other functions that return variations of these event streams for other uses.

nickreynolds: oh you mean if something is delisted but listed again after we've already subscribed to the observable? that could be an issue. I'll remove distinct for now, but there's problems with it either way.

packages/core/src/utils/errors.ts
nickreynolds: I don't think InvalidParameters really makes sense and wouldn't be particularly helpful (user might think the listingAddress they inputted is wrong). I don't think NoChallenge is too granular, we can discuss changing it later if we think it is though.

PR #53, opened 2018-03-25
Abstracted Smart-Contract singleton access in core

ritave:


packages/core/src/artifacts/OwnedAddressTCRWithAppeals.json
```diff
@@ -1183,6 +1183,12 @@
             "events": {},
             "links": {},
             "address": "0x5cf4114912d0b1eacf666e1c6b9fc91eb143956b"
+        },
+        "50": {
```
nickreynolds: is this local ganache network deployment? do we want to commit this if it's local to your machine?

packages/contracts/package.json
```diff
@@ -12,6 +12,7 @@
     "lint": "run-p --aggregate-output lint:*",
     "lint:typescript": "tslint --project ./",
     "lint:solidity": "solium --dir contracts/",
+    "migrate": "truffle migrate",
```
nickreynolds: might want to rename this command 'migrate-mainnet' since it's a "pretty big deal". maybe add another command for 'migrate-rinkeby'?

ritave: It actually throws an error on my computer if no network is specified. I wonder why it works differently on yours

packages/core/src/artifacts/OwnedAddressTCRWithAppeals.json
ritave: This is local ganache network deployment. It's done in a way that if anyone migrates through `yarn ganache`, it's gonna end up as the same address on the same network id

packages/contracts/package.json
nickreynolds: actually you're right. It requires us to specify network because we don't have a default configuration specified in `truffle-config.js`. So as long as we don't define a default configuration we're good.

packages/core/src/artifacts/OwnedAddressTCRWithAppeals.json
nickreynolds: 👍

PR #56, opened 2018-03-26
Adding civil-dapp package

nickreynolds:
Civil-dapp is a React App that allows users to see and interact with Civil contracts without any backend. Acts as a much nicer UX than debug-ui but may not be able to show articles and other information stored on CMS.


ritave:
@nickreynolds also, add the package to README with a small explanation

packages/dapp/src/App.css
```diff
@@ -0,0 +1,28 @@
+.App {
```
walfly: can you switch to styled-components for consistency?

PR #61, opened 2018-03-29
add credibility indicators [ch64]

walfly:
to do, make headline into contentedible instead of input


walfly:
snapshots kind of make it crazy

packages/editor/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@joincivil/editor",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
ritave: Please don't update versions manually. Every once in a while (sprint?) I do a release of all packages through `lerna` that tags the commit all in one place.

If you need to upgrade this package, use lerna which add's tags etc.

packages/editor/src/Editor.stories.tsx
```diff
@@ -17,6 +17,13 @@ import { linkJson } from "./storyFixtures/link";
 import { pullQuoteJson } from "./storyFixtures/pullquote";
 import { imageJson } from "./storyFixtures/image";
 import { imageBreakoutJson } from "./storyFixtures/imagebreakout";
+import {
```
ritave: Should import and add them through a loop somehow?

walfly: There's only ever going to be 4

packages/editor/src/plugins/components/CredibilityIndicators.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
+// @ts-ignore
```
ritave: If it's a linter error use ```// tslint:disable-next-line```.

```// @ts-ignore``` is a compiler directive (not linter) and this one disables the whole compiler reporting errors here, meaning that if styles-components fails to import at all, we're gonna know only at runtime instead of compile time.

@dankins @jorgelo @nickreynolds  : FYI

walfly:
@ritave @juliahimmel this is ready

ritave:
Fix tests and good to go

PR #62, opened 2018-03-29
Add multisig to @joincivil/core

ritave:
TODO:
- [x] Add documentation comments


packages/contracts/contracts/multisig/MultiSigWallet.sol
```diff
@@ -78,15 +78,15 @@ contract MultiSigWallet {
     _;
   }
 
-  modifier notNull(address _address) {
-    require(_address != 0);
+  modifier notNull(address checkedAddress) {
```
nickreynolds: `checkedAddress` and `checkedRequired` below both sound a bit awkward to me. Could `checkRequired` be `requiredConfirmationsProposed`? Or something?

PR #63, opened 2018-03-30
Debug-UI: Parameterizer

nickreynolds:
- Adding Parameterizer info and functionality to Debug-UI
- Adding state to parameterizer proposal wrapper, adding more event stream observables, updating debug-ui to show streams & state
- Adding fancier observables for parameterizer, displaying those in debug-ui. Adding some basic getter functions. Modifying Parameterizer contract slightly to make events more useful.
- adding comments
- cleanup


packages/core/src/artifacts/OwnedAddressTCRWithAppeals.json
```diff
@@ -1136,8 +1136,8 @@
         "4": {
             "events": {},
             "links": {},
-            "address": "0x79dd3705ba094da565806cc2ceef32c72058dc78",
-            "transactionHash": "0xb99b7bf9ae40b1b27845d30e546227bf7e74bcb5b430a07e35cd47e57ac9a297"
+            "address": "0x81cd72f8f3b49f26177f9a1f36eb727f1e27857a",
+            "transactionHash": "0x6c206be49820cab85e06de87fe86bab59d56f91c014fee8cc295a5156c7ba7b8"
         },
         "50": {
```
ritave: What about Ganache, did you deploy the new version there too?

ritave: We're starting to need a better jsons, with versioning. I would love to work on this, especially since 0x has expertize on this, but not enough time for that

packages/core/src/contracts/ownedAddressTCRWithAppeals.ts
```diff
@@ -62,6 +63,12 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
     return Voting.atUntrusted(this.web3Wrapper, await this.getVotingAddress());
   }
 
+  /**
+   * Returns Parameterizer instance associated with this TCR
+   */
+  public async getParameterizer(): Promise<Parameterizer> {
```
ritave: Since the Paramizer's address is in the .json already - you can speed up this process by using the `await ParametizerContract.singleton()` call:
https://github.com/joincivil/Civil/blob/master/packages/core/src/templates/contract/partials/constructor.handlebars#L1

packages/core/src/contracts/parameterizer.ts
```diff
@@ -39,34 +40,240 @@ export class Parameterizer extends BaseWrapper<ParameterizerContract> {
...
+
+  /**
+   * Get address for voting contract used with this TCR
+   */
+  public async getVotingAddress(): Promise<EthAddress> {
+    return this.instance.voting.callAsync();
+  }
+
+  public async getPropState(propID: string): Promise<ParamProposalState> {
+    if (! await this.instance.propExists.callAsync(propID)) {
```
ritave: might want to add parentheses around await

ritave: Also, we can return an enum from the smart-contracts which will be converted to bignum. Might as well return the state with one call here and iterate on that.

packages/core/src/contracts/parameterizer.ts
```diff
@@ -39,34 +40,240 @@ export class Parameterizer extends BaseWrapper<ParameterizerContract> {
...
   /**
    * Event Streams
    */
 
   /**
-   * An unending stream of all active Reparametizations proposed
-   * @param fromBlock Starting block in history for events concerning reparametizations
-   *                  Set to "latest" for only new events
-   * @returns currently active proposals
+   * An unending stream of the propIDs of all active Reparametizations proposed
```
ritave: `An unending stream of IDs of active Reparametizations proposed`?

packages/core/src/contracts/parameterizer.ts
```diff
@@ -39,34 +40,240 @@ export class Parameterizer extends BaseWrapper<ParameterizerContract> {
...
+      .ReparameterizationProposalStream({}, { fromBlock })
+      .map((e) => e.args.propID)
+      .concatFilter(async (propID) => this.isPropInUnchallengedApplicationUpdatePhase(propID));
+
+    const challengesToResolve = this.instance
+      .NewChallengeStream({}, { fromBlock })
+      .map((e) => e.args.propID)
+      .concatFilter(async (propID) => this.isPropInChallengeResolvePhase(propID));
+
+    return Observable.concat(applicationsToProcess, challengesToResolve).distinct();
```
ritave: Use `Observable.merge` here, otherwise you will only get the first stream due to it being unending and concat works in-order on streams

packages/core/src/contracts/parameterizer.ts
```diff
@@ -39,34 +40,240 @@ export class Parameterizer extends BaseWrapper<ParameterizerContract> {
...
+    return Observable.merge(applicationsToProcess, challengesToResolve);
+  }
+
+  /**
+   * An unending stream of the pollIDs of all Reparametization proposals that have
+   * been challenged and had those challenges resolved.
+   * @param fromBlock Starting block in history for events. Set to "latest" for only new events
+   * @returns pollIDs for proposals that have been challenged and resolved
+   */
+  public pollIDsForResolvedChallenges(propID: string, fromBlock: number|"latest" = 0): Observable<BigNumber> {
```
ritave: Do you want to have `export type PollId = BigNumber` for easier documentation?

packages/debug-ui/src/code/parameterizer.ts
```diff
@@ -0,0 +1,68 @@
+import { Civil, ParamProp } from "@joincivil/core";
+import * as marked from "marked";
+import BigNumber from "bignumber.js";
+import { deployNewsroom } from "../../scripts/deploy-newsroom";
+import { proposeReparameterization } from "../../scripts/parameterizerActions";
+
+// Metamask is injected after full load
+window.addEventListener("load", async () => {
```
ritave: Can we have a base startup function?
Something like
```typescript
function initializeDebugUI(callback: (civil: Civil) => void|Promise<void>) {
  window.addEventListener("load", () => {
    setListener();
    callback(new Civil({ debug: true }));
  });
}

PR #67, opened 2018-04-04
Adding header to DApp

nickreynolds:
WIP style props

Adding links but in a really ugly way that I need to fix

cleaning up list items in nav bar


packages/dapp/src/components/Listings.tsx
```diff
@@ -39,19 +40,29 @@ class Listings extends React.Component<{}, ListingsState> {
...
+    let tcr;
+    try {
+      tcr = civil.tcrSingletonTrusted();
+    } catch (ex) {
+      console.log("failed to get tcr.");
+      this.setState({error: "No Supported Network Found. Please set MetaMask network to Rinkeby and Unlock Account."});
+    }
+
+    if (tcr) {
+      const instance = this;
```
walfly: you don't need this with => functions. The arrow means the scope is the same as the outer function

packages/dapp/src/components/navbar/NavBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+  balance: string;
+}
+
+class NavBar extends React.Component<{}, NavBarState> {
+
+  constructor(props: {}) {
+    super(props);
+
+    this.state = {
+      balance: "loading....",
```
walfly: this is fine, but also weird as hell. makes more sense to me to have a loading boolean then set the balance to be loading...

packages/dapp/src/components/navbar/NavBar.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <NavBarItem text="About" link="about"/>
+          <NavBarItem right={true} text={"Your Balance: " + this.state.balance} link="dead"/>
+        </StyledUL>
+    );
+  }
+
+  private initNavBar = () => {
+    const civil = new Civil();
+    const token = civil.getEIP20ForDeployedTCR();
+    const instance = this;
```
walfly: same, not needed

packages/dapp/src/components/navbar/NavBarItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+class NavBarItem extends React.Component<NavBarItemProps> {
+
+  constructor(props: NavBarItemProps) {
+    super(props);
+  }
+
+  public render(): JSX.Element {
+    let listItem: string|JSX.Element = this.props.text;
```
walfly: couldn't just pass the link in as a child of this component

packages/dapp/src/components/navbar/NavBarItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+      listItem = (
+        <Link to={this.props.link} style={{ textDecoration: "none", color: "white"}}>
+          {this.props.text}
+        </Link>
+      );
+    }
+    return (
+      <StyledLI right={this.props.right}>
+          <StyledSpan big={this.props.big}>
+            {listItem}
```
walfly: `{this.props.children}` and remove the logic

packages/dapp/src/components/Listings.tsx
```diff
@@ -39,19 +40,29 @@ class Listings extends React.Component<{}, ListingsState> {
...
-    const instance = this;
-    const subscription = tcr.listingsInApplicationStage().subscribe((listing) => {
-      instance.setState({applications: this.state.applications.push(listing)});
-    });
-    this.setState({applicationSubscription : subscription});
+    let tcr;
+    try {
+      tcr = civil.tcrSingletonTrusted();
+    } catch (ex) {
+      console.log("failed to get tcr.");
```
ritave: For the future - might want to compare what error you got before showing a nice message, and display the error if not recognized

packages/dapp/src/App.tsx
```diff
@@ -1,24 +1,16 @@
...
-
 class App extends React.Component {
   public render(): JSX.Element {
     return (
       <Router>
-      <StyledDiv>
-        <Header />
-        <Main />
-      </StyledDiv>
+        <div>
```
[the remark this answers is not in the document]
nickreynolds: without this div it gives a compiler error saying "A <Router> may have only one child element"

packages/dapp/src/components/navbar/NavBar.tsx
nickreynolds: I'll make sure to do something like that when I do the balance widget "for real"

packages/dapp/src/components/Listings.tsx
```diff
@@ -39,19 +40,29 @@ class Listings extends React.Component<{}, ListingsState> {
     return (
       <StyledDiv>
         whaddup listings: {this.state.applications.toString()}
+        <br/>
+        {this.state.error && this.state.error}
```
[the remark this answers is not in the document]
nickreynolds: this is a low-fi way to say "if this.state.error is defined, render this.state.error". When I build the actual component, this will be done in a nicer way.

https://reactjs.org/docs/conditional-rendering.html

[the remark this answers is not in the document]
walfly: you don't need the double for that if it's undefined or null it won't render anything if its defined it will render itself.

packages/dapp/src/App.tsx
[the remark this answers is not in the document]
walfly: try this syntaxt
```
<>
<Navbar/>
<Main/>
</>
```
it won't add extra html that way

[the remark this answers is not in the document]
nickreynolds: fixed

packages/dapp/src/components/Listings.tsx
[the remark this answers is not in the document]
nickreynolds: that makes sense, fixed

PR #75, opened 2018-04-10
pass value as json to civil editor

walfly:


packages/editor/src/Editor.tsx
```diff
@@ -25,12 +26,22 @@ const StyledEditor = styled(Editor)`
 `;
 
 export class CivilEditor extends React.Component<EditorProps, EditorState> {
+<<<<<<< Updated upstream
```
dankins: is this right? looks like it is a merge conflict

walfly: I thought i got rid of those :(

walfly:
@dankins ready

PR #86, opened 2018-04-11
TCR Refactor - Part 1

nickreynolds:
- rename OwnedAddressTCRWithAppeals to CivilTCR
- adding types for listing, challenge, appeal and modifying tcr wrapper to make fewer async calls (e.g. can just make 1 async call to get Listing and use that to determine state, rather than making an async call in each "isListingInThisState" function)


packages/contracts/contracts/tcr/AddressRegistry.sol
```diff
@@ -37,14 +37,14 @@ contract AddressRegistry {
     bool resolved;          // Indication of if challenge is resolved
     uint stake;             // Number of tokens at stake for either party during challenge
     uint totalTokens;       // (remaining) Number of tokens used in voting by the winning side
-    mapping(address => bool) tokenClaims; // Indicates whether a voter has claimed a reward yet
+    mapping(address => bool) hasClaimedTokens; // Indicates whether a voter has claimed a reward yet
   }
 
   // Maps challengeIDs to associated challenge data
-  mapping(uint => Challenge) internal challenges;
+  mapping(uint => Challenge) public challenges;
```
ritave: Undocumented behaviour use here. Additionally with the Challenge being big and having a mapping inside, it sounds wonky. Does it work?

We could make use of structs long term by supporting abiv2 but it's also a bit of work

packages/contracts/contracts/tcr/CivilTCR.sol
```diff
@@ -21,7 +21,7 @@ contract OwnedAddressTCRWithAppeals is RestrictedAddressRegistry {
   event MakeAppealLengthSet(uint length);
   event AppealLengthSet(uint length);
   event FeeRecipientSet(address recipient);
-
+
```
ritave: whitespace

packages/contracts/migrations/6_registry.ts
```diff
@@ -7,30 +7,23 @@ const Token = artifacts.require("EIP20.sol");
 const DLL = artifacts.require("dll/DLL.sol");
 const AttributeStore = artifacts.require("attrstore/AttributeStore.sol");
 
-const OwnedAddressTCRWithAppeals = artifacts.require("tcr/OwnedAddressTCRWithAppeals.sol");
+const CivilTCR = artifacts.require("tcr/CivilTCR.sol");
```
ritave: while we're at it:
`./tcr/CivilTCR.sol` -> add the relative path to make resolving easier and remove a class of bugs

packages/core/src/artifacts/CivilTCR.json
```diff
@@ -1134,18 +1134,5 @@
...
-      "transactionHash": "0x4a6ca7c6e2d6b639f850750d14fe1c216467099c974acb0978610d27cc7782f7"
-    },
-    "50": {
-      "events": {},
-      "links": {},
-      "address": "0x558b378b0a693c5d4f68589f37121c139420c8a6",
-      "transactionHash": "0x996142214ef54b560472817d327e33821d41c945a360701e0218a88a5ed800eb"
-    }
-  }
+  "networks": {}
```
ritave: Migrate everything?

packages/core/src/civil.ts
```diff
@@ -123,12 +123,12 @@ export class Civil {
...
-    return OwnedAddressTCRWithAppeals.singleton(this.web3Wrapper, this.contentProvider);
+  public tcrSingletonTrusted(): CivilTCR {
+    return CivilTCR.singleton(this.web3Wrapper, this.contentProvider);
   }
 
   /**
-   * Returns the EIP20 instance associated with the deployed OwnedAddressTCRWithAppeals
+   * Returns the EIP20 instance associated with the deployed CivilTCR
    */
   public async getEIP20ForDeployedTCR(): Promise<EIP20> {
```
ritave: by the way, why the helper functions. Might as well get this stuff from the tcr?

packages/contracts/contracts/tcr/AddressRegistry.sol
nickreynolds: the mapping inside the Challenge struct is the only part that's not exposed publicly. When you do `registry.challenges(challengeID)` it will return an array with everything but the `hasClaimedTokens` mapping, so we still need the getter function for that.

packages/debug-ui/src/code/newsroomEntry.ts
```diff
@@ -17,7 +17,15 @@ initializeDebugUI(async civil => {
   const displayState = document.getElementById("state")!;
   displayState.innerHTML = "State: " + ListingState[state].toString();
 
-  const challengeId = await tcr.getListingChallengeID(address);
+  const listing = await tcr.getListing(address);
+  const appeal = await tcr.getAppeal(address);
+
+  const newsroom = await civil.newsroomAtUntrusted(address);
+  const owners = await newsroom.owners();
+  const owner = await newsroom.owner();
```
ritave: Use just owners

packages/core/src/contracts/newsroom.ts
```diff
@@ -94,6 +108,10 @@ export class Newsroom extends BaseWrapper<NewsroomContract> {
     return this.multisigProxy.owners();
   }
 
+  public async owner(): Promise<EthAddress> {
```
ritave: Bad idea. We might have some different way of ACL on Newsroom in the future. Using just owners abstracts it away.

If you wanna check for multisig, add `hasBoardOfDirectors(): Promise<EthAddress?>` or something like this and add this functionallity to `BaseMultisigProxy` (`return isDefined(this.multisig);`);

packages/core/src/civil.ts
```diff
@@ -80,14 +78,25 @@ export class Civil {
...
     return Newsroom.deployTrusted(this.web3Wrapper, this.contentProvider, newsroomName);
   }
 
+  /**
+   * Create a new Newsroom which is not owned by a multisig on the current Ethereum network with the
+   * bytecode included in this library
+   * The smart contract is trusted since it comes from a trusted source (us).
+   * This call may require user input - such as approving a transaction in Metamask
+   */
+  public async newsroomDeployNonMultisigTrusted(newsroomName: string): Promise<TwoStepEthTransaction<Newsroom>> {
```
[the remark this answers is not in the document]
nickreynolds: well I'm using it right now in debug ui, since the multisig version wasn't working when trying to apply to TCR (something we need to fix in future PR). I think it will be useful to have both of these for testing purposes.

packages/contracts/migrations/2_optional_for_test.ts
```diff
@@ -5,7 +5,7 @@ import BN from "bignumber.js";
 import { config } from "./utils";
 import { MAIN_NETWORK } from "./utils/consts";
 
-const Token = artifacts.require("tokens/eip20/EIP20.sol");
+const Token = artifacts.require("EIP20");
```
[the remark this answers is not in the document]
nickreynolds: I noticed that a few were already set up like this, so I made the change and it seems to work. The artifacts are all in the same folder, so it makes sense that we don't need the paths.

packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -547,17 +617,32 @@ export class OwnedAddressTCRWithAppeals extends BaseWrapper<OwnedAddressTCRWithA
...
+    if (!l.challengeID.isZero()) {
+      const voting = await this.getVoting();
+      const revealPeriodExpiryTimestamp = await voting.getRevealPeriodExpiry(l.challengeID);
       const revealPeriodExpiry = new Date(revealPeriodExpiryTimestamp.toNumber() * 1000);
       if (revealPeriodExpiry < new Date()) {
-        const isPassed = await voting.isPollPassed(challengeID);
+        const isPassed = await voting.isPollPassed(l.challengeID);
         if (!isPassed) {
-          const requestAppealPhaseExpiry = await this.getRequestAppealExpiryTimestamp(listingAddress);
+          let a: Appeal;
```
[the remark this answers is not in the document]
nickreynolds: if it's at the top of the function, then it's making an async call that might be unnecessary. I'd rather make that call only when we know it makes sense to do so

packages/core/src/contracts/newsroom.ts
```diff
@@ -62,6 +62,20 @@ export class Newsroom extends BaseWrapper<NewsroomContract> {
...
+  public static async deployNonMultisigTrusted(
+    web3Wrapper: Web3Wrapper,
+    contentProvider: ContentProvider,
+    newsroomName: string,
+  ): Promise<TwoStepEthTransaction<Newsroom>> {
+    const txData: TxData = { from: web3Wrapper.account };
+    return createTwoStepTransaction(
+      web3Wrapper,
+      await NewsroomContract.deployTrusted.sendTransactionAsync(web3Wrapper, newsroomName, txData),
+      async receipt => this.atUntrusted(web3Wrapper, contentProvider, receipt.contractAddress!),
```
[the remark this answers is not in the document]
ritave: @nickreynolds ping, simple change?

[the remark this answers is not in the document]
nickreynolds: sure I'll fix that up & merge

PR #90, opened 2018-04-13
refactor contract to let editors just publish stuff

walfly:


packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -2,18 +2,14 @@ pragma solidity ^0.4.19;
...
   event NameChanged(string newName);
 
   string private constant ROLE_REPORTER = "reporter";
   string private constant ROLE_EDITOR = "editor";
 
   uint private latestId;
-  mapping(uint => Content) private content;
-  mapping(uint => bool) private waiting;
-  mapping(uint => bool) private approved;
+  mapping(uint => Revision) private content;
```
nickreynolds: we can make this mapping public, actually, and get rid of the getters (see PR 86 changes to Listing/Challenge/Appeal mappings)

packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -2,18 +2,14 @@ pragma solidity ^0.4.19;
 import "./ACL.sol";
 
 contract Newsroom is ACL {
-  event ContentProposed(address indexed author, uint indexed id);
-  event ContentApproved(uint id);
-  event ContentDenied(uint id);
+  event ContentAdded(address indexed editor, uint indexed id);
```
nickreynolds: Added -> Published might be more clear? Also might be worth it to add a bit more info to this event (url would be helpful)

nickreynolds: RevisionPublished? Then we can have a separate event that fires in addition to this one if the article is brand new, as opposed to just a revision

packages/contracts/test/newsroom/newsroom.ts
```diff
@@ -11,6 +11,7 @@ const expect = chai.expect;
 
 const FIRST_NEWSROOM_NAME = "TEST NAME, PLEASE IGNORE";
 const SOME_URI = "http://thiistest.uri";
+const SOME_HASH = web3.sha3();
```
nickreynolds: for this you should be able to just do `const SOME_HASH = "0x01"`

packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -56,41 +32,28 @@ contract Newsroom is ACL {
     _removeRole(who, role);
   }
 
-  function proposeContent(string contentUri) public requireRole(ROLE_REPORTER) returns (uint) {
+  function addContent(string contentUri, bytes32 contentHash) public requireRole(ROLE_EDITOR) returns (uint) {
```
nickreynolds: publishContent() feels better

packages/core/src/civil.ts
```diff
@@ -14,7 +14,7 @@ const debug = Debug("civil:main");
 
 export interface CivilOptions {
   web3Provider?: Web3.Provider;
-  contentProvider?: ContentProvider;
+  ContentProvider?: ContentProviderCreator;
```
nickreynolds: why is ContentProvider capitalized here?

packages/core/src/content/ipfsprovider.ts
```diff
@@ -29,10 +29,10 @@ export class IPFSProvider implements ContentProvider {
     });
   }
 
-  public async put(content: string): Promise<Uri> {
-    return new Promise<Uri>((resolve, reject) => {
+  public async put(content: string): Promise<ContentHeader> {
+    return new Promise<ContentHeader>((resolve, reject) => {
       ipfs.add(Buffer.from(content), (err: any, ipfsHash: any) => {
-        resolve(this.scheme() + "://" + ipfsHash[0].path);
+        resolve({ uri: this.scheme() + "://" + ipfsHash[0].path, hash: ipfsHash });
```
nickreynolds: I think ipfsHash here is actually poorly named. ipfsHash[0].path gets the actual hash, I believe

packages/core/src/civil.ts
walfly: its a Class definition not an instance

walfly:
Wooooo

PR #93, opened 2018-04-17
Start Listing Component in DApp

nickreynolds:
- subscribing to more event streams in DApp index, starting on Listing component which shows history of listing
- small modifications to core to support the above

### DIFF
--- a/packages/core/src/contracts/tcr/civilTCR.ts
+++ b/packages/core/src/contracts/tcr/civilTCR.ts
@@ -6,11 +6,11 @@ import "@joincivil/utils";
 import { Voting } from "./voting";
 import { Parameterizer } from "./parameterizer";
 import { BaseWrapper } from "../basewrapper";
-import { CivilTCRContract } from "../generated/civil_t_c_r";
+import { ApplicationArgs, CivilTCRContract } from "../generated/civil_t_c_r";
 import { Web3Wrapper } from "../../utils/web3wrapper";
 import { ContentProvider } from "../../content/contentprovider";
 import { CivilErrors, requireAccount } from "../../utils/errors";
-import { Appeal, EthAddress, Listing, ListingState, TwoStepEthTransaction, Challenge } from "../../types";
+import { Appeal, EthAddress, Listing, ListingState, TwoStepEthTransaction, Challenge, WrappedEvent } from "../../types";
 import { createTwoStepSimple, is0x0Address, isEthAddress } from "../utils/contracts";
 import { EIP20 } from "./eip20";
 
@@ -115,6 +115,15 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
       .ApplicationStream({}, { fromBlock })
       .map(e => e.args.listingAddress)
       .concatFilter(async listingAddress => this.isInApplicationPhase(listingAddress));
+  }
+
+  public listingApplications(listingAddress: EthAddress): Observable<Promise<WrappedEvent<ApplicationArgs>>> {
+    return this.instance.ApplicationStream({ listingAddress }, { fromBlock: 0 }).map(async e => {
+      return {
+        decodedLogEntryEvent: e,
+        timestamp: await this.web3Wrapper.getTimestampForBlock(e.blockNumber!),
+      };
+    });
   }
 
   /**
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -79,7 +79,7 @@ export enum AbiType {
 }
 
 export { CivilEventArgs, Artifact } from "./contracts/generated/artifacts";
-
+export { ApplicationArgs } from "./contracts/generated/civil_t_c_r";
 export interface DecodedTransactionReceipt<T> {
   blockHash: string;
   blockNumber: number;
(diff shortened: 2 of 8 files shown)


packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -6,11 +6,11 @@ import "@joincivil/utils";
 import { Voting } from "./voting";
 import { Parameterizer } from "./parameterizer";
 import { BaseWrapper } from "../basewrapper";
-import { CivilTCRContract } from "../generated/civil_t_c_r";
+import { ApplicationArgs, CivilTCRContract } from "../generated/civil_t_c_r";
```
ritave: FYI: Using Args was removed in master. You might want to use ApplicationLog

packages/core/src/types.ts
```diff
@@ -175,3 +175,8 @@ export interface ParamProp {
   proposedValue: BigNumber;
   pollID?: BigNumber;
 }
+
+export interface WrappedEvent<T> {
```
ritave: how about this:

```typescript
export interface TimestampedEvent<T extends DecodedLogEntryEvent> extends T {
  timestamp: () => Promise<BigNumber>;
}
export function createTimestamped<T extends DecodedLogEntryEvent>(web3Wrapper: Web3Wrapper, event: T) {
  return {
    ...event,
    timestamp: async () = (await web3Wrapper.getBlock(e.blockNumber!)).timestamp 
  }
}

public listingApplications(listingAddress: EthAddress): Observable<TimestampedEvent<ApplicationLog>> {
  return this
    .instance
    .ApplicationStream({ listingAddress }, { fromBlock: 0 })
    .map(createTimestamped<ApplicationLog>.bind(null, web3Wrapper));
}
```

packages/core/src/utils/web3wrapper.ts
```diff
@@ -71,6 +71,14 @@ export class Web3Wrapper {
     return this.web3.version.network;
   }
 
+  public async getTimestampForBlock(blockNumber: number): Promise<BigNumber> {
+    return new Promise<BigNumber>((resolve, reject) => {
```
ritave: I think this is too narrow, since they can just get the timestamp after getting block.

How about:

```typescript
public async getBlock(blockNumber: number): Promise<Web3.Block> {
  const getBlockAsync = promisify<Web3.Block>(this.web3.eth.getBlock, this.web3.eth);
  return getBlockAsync(blockNumber);
}
```

packages/core/src/types.ts
ritave: If you REALLY want to have it resolved from beginning use:

```typescript
public listingApplications(listingAddress: EthAddress): Observable<TimestampedEvent<ApplicationLog>> {
  return this
    .instance
    .ApplicationStream({ listingAddress }, { fromBlock: 0 })
    .flatMap(async e => { // notice the flatMap
      ...e,
      timestamp: (await this.web3.getBlock(e.blockNumber!)).timestamp;
    });
```

packages/dapp/src/components/ListingHistory.tsx
```diff
@@ -0,0 +1,131 @@
...
+        {this.state.listingHistory.map(e => {
+          return <ListingEvent key={this.props.match.params.listing + e.blockNumber} event={e} />;
+        })}
+      </StyledDiv>
+    );
+  }
+
+  // TODO(nickreynolds): move this all into redux
+  private initListings = async () => {
+    const civil = new Civil();
```
ritave: Hum, this also might fail when there's not metamask. Ought to make the constructor safe

packages/dapp/src/components/Listings.tsx
```diff
@@ -13,6 +16,20 @@ const StyledDiv = styled.div`
 export interface ListingsState {
   applications: List<string>;
   applicationSubscription: Subscription;
+  whitelistedListings: List<string>;
+  whitelistSubscription: Subscription;
```
ritave:
```typescript
compositeSubscription: Subscription
...
this.state.compositeSubscription = new Subscription();
this.state.composieSubscription.
  .add(tcr.currentChallengedRevealVotePhaseListings().subscribe(listing => 
    this.setState({ inChallengeRevealListings: this.state.inChallengeRevealListings.push(listing) }));
this.state.composieSubscription.
  .add(tcr.(...).subscribe(...));
...
this.state.compositeSubscribption.unsubscribe(); // everything at once
```

packages/dapp/src/components/Listings.tsx
```diff
@@ -58,10 +119,45 @@ class Listings extends React.Component<{}, ListingsState> {
     }
 
     if (tcr) {
-      const subscription = tcr.listingsInApplicationStage().subscribe(listing => {
+      const applicationSubscription = tcr.listingsInApplicationStage().subscribe(listing => {
         this.setState({ applications: this.state.applications.push(listing) });
       });
-      this.setState({ applicationSubscription: subscription });
+      this.setState({ applicationSubscription });
```
ritave: maybe this?

```typescript
const stateSubscribe = (key:string) => ((event: any) => this.setState({`${key}`: this.state[key].push(event) }))

this.state.compositeSubscription.add(tcr.whitelistedListings().subscribe(stateSubscribe("whitelistedListings"));
this.state.compositeSubscription.add(tcr.readyToBeWhitelistedListings().subscribe(stateSubscribe("readyToWhitelistListings"));
```

nickreynolds:
closing without merge, setting up new PR later today

PR #94, opened 2018-04-18
Create our own typescript typings structure

ritave:


packages/core/src/types.ts
```diff
@@ -78,9 +80,9 @@ export enum AbiType {
   Fallback = "fallback",
 }
 
-export { CivilEventArgs, Artifact } from "./contracts/generated/artifacts";
+export { CivilLogs, Artifact } from "./contracts/generated/artifacts";
 
-export interface DecodedTransactionReceipt<T> {
+export interface DecodedTransactionReceipt<L extends DecodedLogEntry> {
```
nickreynolds: is it standard to use "L" here if it's being extended? or just personal choice?

ritave: It's a personal choice. Naming it `L` as in `Log`

nickreynolds: 👍

PR #98, opened 2018-04-19
fixing npm badge links with dashes

godfreyhobbs:
Dashes --	→	- Dash


nickreynolds:
in order to fix the lint test, just run "yarn prettier" from root of monorepo and commit the changes to the README. Better than trying to fix it manually (if that's what you're doing)

godfreyhobbs:
So the prettier issue is due to the updated version of prettier being released - prettier@1.12.1.

It is not related to the `README.md` changes.  Do you want me to commit the prettier fixes?

Also NOT sure why you are running prettier as part of your CI?  Why not just use a standard `lint`.

walfly:
Why have prettier if you're not going to use it in your ci?

walfly:
we should probably ignore readme files though

dankins:
@godfreyhobbs Thank you so much for your contribution! Feel free to reach out if you have any questions or if there is any areas of the codebase you are particularly interested in. We're still trying to bootstrap our dev community so keep an eye on this space.

godfreyhobbs:
@walfly readme is not the issue.  The problem is that prettier is not stable.   So At least, lock down the version of prettier

godfreyhobbs:
This is why the build failed and why prettier needs to be pinned.

```javascript   * @param propID ID of prop to check
    */
   public async isPropInChallengeRevealPhase(propID: string): Promise<boolean> {
-    if (!await this.instance.propExists.callAsync(propID)) {
+    if (!(await this.instance.propExists.callAsync(propID))) {
       return false;
     }
     const pollID = await this.instance.getPropChallengeID.callAsync(propID);
```

walfly:
hmm, i greatly prefer the change it's asking for though.  more clear that the `!` operates on the value of the expression

godfreyhobbs:
I originally updated a readme.  Do you want me to also check in the prettier code change?

walfly:
nah, its fine, thanks

PR #104, opened 2018-04-24
Added basic solidity coverage

ritave:
This is a bit hacky. Long-term solution would be include support for truffle artifacts in 0x's solcov


nickreynolds:
can we get a badge added that displays the solidity coverage %?

ritave:
@nickreynolds sounds good, I'll add it when I'll get the rest working and see first results on coveralls

PR #106, opened 2018-04-25
using deployed token for rinkeby migrations

nickreynolds:
instead of deploying a new one every time

### DIFF
--- a/packages/contracts/conf/config.json
+++ b/packages/contracts/conf/config.json
@@ -20,7 +20,7 @@
     "appealChallengeCommitStageLength": 160,
     "appealChallengeRevealStageLength": 160
   },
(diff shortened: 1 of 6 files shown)


packages/contracts/conf/config.json
```diff
@@ -20,7 +20,7 @@
     "appealChallengeCommitStageLength": 160,
     "appealChallengeRevealStageLength": 160
   },
-  "TokenAddress": "0x337cDDa6D41A327c5ad456166CCB781a9722AFf9",
+  "rinkebyTokenAddress": "0x617e6d6db50b2697b1fa37f43ec2834648c39020",
```
ritave: Put it into `testnets.rinkeby`?

PR #112, opened 2018-04-26
Coverage for `setAppealFee`, `setMakeAppealLength`, & `setAppealLength`

petersng:
Some additional small tests for CivilTCR.sol.  Let me know if it makes sense to move these into another file, but it seemed to go with appellate specific grouping in CivilTCR.sol.

Ticket: http://app.clubhouse.io/civil/story/635/


packages/contracts/test/tcr/registryWithAppeals/grantAppeal.ts
```diff
@@ -13,6 +13,87 @@ const PLCRVoting = artifacts.require("PLCRVoting");
 const NEWSROOM_NAME = "unused newsroom name";
 
 contract("Registry With Appeals", accounts => {
+  describe("Function: setAppealFee", () => {
```
nickreynolds: yeah I think these should each go into their own file. 1 file per function that you're testing, basically. Besides that, they look great and I'll approve it when that's done.

PR #114, opened 2018-04-26
Added abillity to post signed revision to smart contract

ritave:


ritave:
@walfly I didn't touch multiple revisions, one feature add a time. I'll add signing first

PR #115, opened 2018-04-27
Adding Government Contract

nickreynolds:
- Separating some appeals functionality into its own contract so that a government can update the appeals entity on its own.
- Slightly reduces gas cost of CivilTCR contract deployment.
- Temporarily commenting out `example.ts` test since it is failing even though I didn't modify `Newsroom` or `NewsroomFactory` at all.


packages/contracts/contracts/tcr/CivilTCR.sol
```diff
@@ -61,22 +56,16 @@ contract CivilTCR is RestrictedAddressRegistry {
   @param tokenAddr          Address of the TCR's intrinsic ERC20 token
   @param plcrAddr           Address of a PLCR voting contract for the provided token
   @param paramsAddr         Address of a Parameterizer contract 
-  @param appellateAddr      Address of appellate entity, which could be a regular user, although multisig is recommended
+  @param govtAddr             Address of a IGovernment contract
```
petersng: Minor, looks like the spacing is off here.

packages/contracts/test/tcr/government/setAppellate.ts
```diff
@@ -0,0 +1,33 @@
...
+      government = await Government.at(await registry.government());
+    });
+
+    it("should be possible for JAB to set appellate", async () => {
+      await expect(government.setAppellate(newAppellate, { from: JAB })).to.eventually.be.fulfilled(
+        "Should have allowed JAB to update appellate",
+      );
+    });
+
+    it("should not be possible for troll to set judgeAppealLength value", async () => {
```
petersng: Minor, but the string on this should be `should not be possible for troll to set appellate`.

packages/contracts/test/tcr/government/get.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    it("judgeAppealPhaseLength value should be same as in config after deployment", async () => {
+      const judgeAppealLength = await government.get("judgeAppealPhaseLength");
+      expect(judgeAppealLength).to.be.bignumber.equal(
+        new BigNumber(utils.paramConfig.judgeAppealPhaseLength),
+        "judgeAppealPhaseLength was not equal to value in config immediately after deployment",
+      );
+    });
+
+    it("should not be possible for troll to set judgeAppealPhaseLength value", async () => {
```
petersng: The test here doesn't appear to match the description.  Prob don't need to test for trolls since we test it in the `set()` tests.

packages/contracts/contracts/interfaces/IGovernment.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.4.19;
+
+/**
+@title IGovernemnt
```
petersng: Minor typo

packages/contracts/conf/config.json
```diff
@@ -2,27 +2,30 @@
   "paramDefaults": {
     "minDeposit": 10,
     "pMinDeposit": 100,
-    "applyStageLength": 120,
+    "applyStageLength": 18000,
```
petersng: Just curious, what is the reason for the bump in these values?

packages/contracts/contracts/interfaces/IGovernment.sol
```diff
@@ -0,0 +1,13 @@
...
+@notice This is an interface that defines the functionality required by a Government
+The functions herein are accessed by the CivilTCR contract as part of the appeals process.
+@author Nick Reynolds - nick@joincivil.com
+*/
+interface IGovernment {
+  function getAppellate() public view returns (address);
+  function getGovernmentController() public view returns (address);
+  function get(string name) public view returns (uint);
```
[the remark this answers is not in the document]
nickreynolds: going to leave it for now. `get` is what the Parameterizer uses, so it seems somewhat conventional.

packages/contracts/conf/config.json
nickreynolds: these are just test values for now, 120 was way too short to do real testing so I bumped up the values (by doing reparameterization proposals) and changed values here so they don't get reverted when we deploy a new version

PR #116, opened 2018-04-27
DApp: Token Holder can Commit Vote on a Challenge

am9u:


packages/dapp/src/components/listing/CommitVoteDetail.tsx
```diff
@@ -0,0 +1,155 @@
...
+export interface InputElementProps {
+  type: string;
+  name?: string;
+  value?: string;
+  readOnly?: boolean;
+  onChange?(event: any): void;
+  validate?(event: any): boolean;
+}
+
+class InputElement extends React.Component<InputElementProps> {
```
nickreynolds: should probably pull this out into `utility`?

packages/dapp/src/components/listing/CommitVoteDetail.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    this.updateCommitVoteParam = this.updateCommitVoteParam.bind(this);
+  }
+
+  public render(): JSX.Element {
+    return (
+      <StyledFormContainer>
+        <h3>Vote On Challenge</h3>
+
+        {/* @TODO/jon: Change this to a hidden form field.
```
nickreynolds: can you change "@TODO/jon" to "TODO(am9u):" ? easier to search for TODOs by person if we're all using the same format

packages/dapp/src/components/listing/CommitVoteDetail.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  private updateCommitVoteParam(event: any): void {
+    const paramName = event.target.getAttribute("name");
+    const val = event.target.value;
+    const newState = {};
+    newState[paramName] = val;
+    this.setState(newState);
+  }
+
+  // @TODO/jon: Make a nicer validation check than this. But for
```
nickreynolds: same here

packages/dapp/src/components/listing/CommitVoteDetail.tsx
```diff
@@ -0,0 +1,155 @@
...
+            secondTransaction={this.commitVoteOnChallenge}
+          >
+            Commit Vote
+          </TransactionButton>
+        </FormGroup>
+      </StyledFormContainer>
+    );
+  }
+
+  private updateCommitVoteParam(event: any): void {
```
nickreynolds: if you change this to `updateCommitVoteParam = (event:any): void => {` you can remove the `bind` in the constructor

packages/dapp/src/apis/civilTCR.ts
```diff
@@ -58,3 +59,34 @@ export async function getNewsroom(address: EthAddress): Promise<any> {
...
+export async function requestVotingRights(numTokens: BigNumber): Promise<TwoStepEthTransaction> {
+  const civil = getCivil();
+  const tcr = civil.tcrSingletonTrusted();
+
+  const voting = tcr.getVoting();
+  const eip = await tcr.getToken();
+
+  const approvedTokensForSpender = await eip.getApprovedTokensForSpender(voting.address);
+  if (approvedTokensForSpender < numTokens) {
+    const approveSpender = await eip.approveSpender(voting.address, numTokens);
```
walfly:
```
const { awaitReceipt } = await eip.approveSpender(voting.address, numTokens);
await awaitReceipt();
```
maybe we should just name it `receipt` 🤔

packages/dapp/src/components/listing/CommitVoteDetail.tsx
```diff
@@ -0,0 +1,106 @@
...
+      isVoteTokenAmtValid: false,
+    };
+  }
+
+  public render(): JSX.Element {
+    return (
+      <StyledFormContainer>
+        <h3>Vote On Challenge</h3>
+
+        {/* @TODO(jon): Change this to a hidden form field.
```
walfly: in the dapp there's no way that you'd ever submit formdata so there's no reason to store extra data in the DOM.  If the user doesn't need to see it, you can just keep it in state or props or redux when we get that setup.

packages/dapp/src/components/utility/FormElements.tsx
```diff
@@ -0,0 +1,52 @@
+import * as React from "react";
+import styled from "styled-components";
+
+export const StyledFormContainer = styled.div`
+  margin: 1em 0;
```
walfly: why ems?

packages/dapp/src/components/utility/FormElements.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public render(): JSX.Element {
+    return (
+      <input
+        type={this.props.type}
+        name={this.props.name}
+        value={this.props.value}
+        readOnly={this.props.readOnly}
+        onChange={this.onChange.bind(this)}
```
walfly: lets be consistent with how we bind in components.  used the `private boundFunction = () => {}` above

packages/dapp/src/components/utility/FormElements.tsx
am9u: No particular reason, I just wanted to give the component a little room to breathen so I added something quick w/o real thought behind it. 

I figure we'll be updating these when we add actual styles/design and can use `rem` or `px` at that point.

packages/dapp/src/components/utility/FormElements.tsx
am9u: Good catch

packages/dapp/src/components/listing/CommitVoteDetail.tsx
am9u: Doh! Very good point. I'll change the comment but leave the field in now for debugging's sake.

PR #120, opened 2018-04-29
Added Article signing ability to the core

ritave:


ritave:
@walfly Hehe, to test this branch https://coveralls.io/builds/16759901/source?filename=packages%2Fcontracts%2Fcontracts%2Fnewsroom%2FNewsroom.sol#L70 I would have to crack private-key encryption and find the private key to the 0x0 address

PR #121, opened 2018-04-29
Removed namespaces

ritave:


packages/core/src/templates/contract/partials/event_types.handlebars
```diff
@@ -1,41 +1,31 @@
...
+{{/each}}
 
-    export type All =
-    {{#each events}}
-      Logs.{{name}}{{#unless @last}} |{{/unless}}{{^unless @last}};{{/unless}}
-    {{/each}}
-  }
+export type {{contractName}}LogsAll =
+{{#each events}}
+  {{../contractName}}Logs{{name}}{{#unless @last}} |{{/unless}}{{^unless @last}};{{/unless}}
```
walfly: what does the `../` do?

walfly: found it in the docs.

PR #122, opened 2018-04-30
Added Event Storage to Core

ritave:


walfly:
This is different than how I was expecting it.  In this set up a newsroom would push content to the event log and add it as a new revision.  I thought that we didn't want the contract to point to the copy in event storage.  I thought you would push the revision into the event storage and index it by the newsroom and its content id so that you could find it easily if you were looking but it wouldn't be pointed to in the contract.  

I'll have to update https://drive.google.com/file/d/1kwCFZlDJRL_tqaxBPQ4O3ZtfuZuwthMJ/view?usp=sharing

walfly:
In the case of a retraction do we have to remove that url from the contract(not totally remove it, but remove it from all future blocks)

packages/contracts/contracts/newsroom/EventStorage.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.4.19;
+
+contract EventStorage {
```
nickreynolds: I thought this functionality would be part of the Newsroom contract. The editor would enter the articleId, revisonId, and the string.

packages/contracts/contracts/newsroom/EventStorage.sol
```diff
@@ -0,0 +1,12 @@
...
+  event StringStored(uint indexed id, string data);
+
+  uint private latestId;
+
+  function store(string data) public {
```
nickreynolds: I don't think this function should increment latestId. An editor might want to split up the article posting into multiple transactions. Should just emit an event (articleId, revisionId, content)

packages/contracts/contracts/newsroom/EventStorage.sol
ritave: EventStorage was discussed as a temporary solution. This is done on purpose to avoid permanent hacks in the Newsroom contract, as well as to avoid bloat in the Newsroom contract and face problems similar to CivilTCR.

To make it into one transaction, one could create a quick smart-contract wrapper which archives as well as publishes a new revision in one transaction

PR #123, opened 2018-04-30
dApp: Countdown timer improvments

am9u:
This update displays the phase countdowns in the dApp in a more human-friendly format (days, months, hours).

It also adds the countdown timer to the `Commit Vote to Challenge` view.


packages/utils/src/viewHelpers.ts
```diff
@@ -0,0 +1,22 @@
...
+// Accepts `seconds` as a single argument and returns a
+// human-readable string of a duration of time.
+// ie: 1 day, 3 hours, 4 minutes, 20 seconds
+export function getReadableDuration(seconds: number): string {
+  const out = [];
+  let period: [number, string];
+  const periods: Array<[number, string]> = [[86400, "day"], [3600, "hour"], [60, "minute"], [1, "second"]];
+  let secondsRemaining: number = seconds;
+
+  while (periods.length) {
```
walfly: you could simplify this with reduce.

```
let secs: number = this.state.secondsRemaining;
return periods.reduce((acc: string, item: [number, string], index: number) => {
   const periodUnits = Math.floor(secs / item[0]);
   if (periodUnits > 0) {
      acc += `${periodUnits} ${item[1]}${periodUnits !== 1 ? "s" : ""}${index < periods.length -1 ? " ": ""}`;
      secs = secs % item[0];
   }
   return acc;
}, "");
```

am9u: I forgot about template literals being available. This is a good suggestion.. I'm going to refactor my code.

walfly: you can use almost everything

PR #125, opened 2018-05-01
Fix namespace, namespaces are fine (i was looking at old issues).  Problem is unused enums get stripped

walfly:
This fixes errors in dapp

### DIFF
--- a/packages/core/src/contracts/multisig/basemultisigproxy.ts
+++ b/packages/core/src/contracts/multisig/basemultisigproxy.ts
@@ -1,5 +1,4 @@
 import { isDeployedBytecodeEqual, isDefined } from "@joincivil/utils";
-import { DecodedLogEntry } from "@joincivil/typescript-types";
 
 import { OwnableContract } from "../interfaces/ownable";
 import { Multisig } from "./multisig";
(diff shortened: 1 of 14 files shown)


packages/core/src/contracts/multisig/basemultisigproxy.ts
```diff
@@ -56,9 +57,9 @@ export class BaseMultisigProxy {
       isProxied: true,
       proxiedId: async () => {
         const receipt = await twoStep.awaitReceipt();
-        const submissionLogs = receipt.logs
-          .filter(log => isDecodedLog(log) && log.event === MultiSigWalletEvents.Submission)
-          .map(log => log as MultiSigWalletLogsSubmission);
+        const submissionLogs: Array<DecodedLogEntry<MultiSigWallet.Args.Submission>> = receipt.logs
```
ritave:
```typescript
const submissionLogs = receipt.logs
  .filter(log => isDecodedLog(log) && log.event === MultiSigWallet.Events.Submission)
  .map(log => log as MultiSigWallet.Logs.Submission);
```

packages/core/src/templates/contract/partials/event_types.handlebars
```diff
@@ -1,31 +1,41 @@
-export const enum {{contractName}}Events {
-  {{#each events}}
-  {{this.name}} = "{{this.name}}",
-  {{/each}}
-}
+// tslint:disable:no-namespace
+export namespace {{contractName}} {
+  export const enum Events {
```
ritave: Instead of forcing existence of all `const enums`, removing `const` modifier from here might be less hacky

walfly: its weird that there is a distinction between enum and const enum

PR #131, opened 2018-05-02
refactoring TCR contracts to remove updateStatus transaction

nickreynolds:
- this saves on the gas cost of deploying contracts. 
- Updating unit tests based on changes.


walfly:
why you close?

nickreynolds:
some build stuff tricked me into thinking I had lowered gas cost when in reality I had increased it, so this PR was basically the opposite of what I was trying to do

PR #136, opened 2018-05-02
Sruddy/ch638/dapp parameterizer view

sruddy:


sruddy:
@nickreynolds sure I will add that link

packages/dapp/src/components/Parameterizer.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <div><StyledSpan>pProcessBy:</StyledSpan> {this.state.pProcessBy}</div>
+        <div><StyledSpan>challengeAppealLen:</StyledSpan> {this.state.challengeAppealLen}</div>
+        <div><StyledSpan>challengeAppealCommitLen:</StyledSpan> {this.state.challengeAppealCommitLen}</div>
+        <div><StyledSpan>challengeAppealRevealLen:</StyledSpan> {this.state.challengeAppealRevealLen}</div>
+      </>
+    );
+  }
+
+  private initParameterValues = async (): Promise<any> => {
+    Object.keys(this.state).forEach(async parameterKey => {
```
walfly: should we do this all at once instead of batching? like pass in `Object.keys(this.state)`, get back and object with a value for every key call setState once.

PR #145, opened 2018-05-03
tcr pre-audit fixes

nickreynolds:
- making updates to TCR contracts (PLCRVoting, Parameterizer) based on latest versions of those contracts available from Consensys and AdChain respectively
- updating DLL to be equivalent to 1.0.4 (can't use EthPM due to doxity issues)
- fixing unit tests based on updates

### DIFF
--- a/packages/contracts/contracts/tcr/PLCRVoting.sol
+++ b/packages/contracts/contracts/tcr/PLCRVoting.sol
@@ -1,8 +1,9 @@
 pragma solidity ^0.4.19;
 
-import "./installed_contracts/tokens/contracts/eip20/EIP20.sol";
+import "./installed_contracts/tokens/contracts/eip20/EIP20Interface.sol";
 import "./installed_contracts/attrstore/contracts/AttributeStore.sol";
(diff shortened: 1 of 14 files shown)


packages/contracts/contracts/tcr/PLCRVoting.sol
```diff
@@ -14,25 +15,29 @@ contract PLCRVoting {
   // EVENTS:
   // ============
 
-  event VoteCommitted(address indexed voter, uint indexed pollID, uint numTokens);
-  event VoteRevealed(address indexed voter, uint indexed pollID, uint numTokens, uint choice);
-  event PollCreated(uint voteQuorum, uint commitDuration, uint revealDuration, uint indexed pollID);
-  event VotingRightsGranted(address indexed voter, uint numTokens);
-  event VotingRightsWithdrawn(address indexed voter, uint numTokens);
+  event _VoteCommitted(uint indexed pollID, uint numTokens, address indexed voter);
```
dankins: is the underscore here just to be consistent with upstream, or for some other reason?

PR #146, opened 2018-05-04
DApp User can see Listings that are kept [ch634]

walfly:


walfly:
still finishing actions part of it

packages/dapp/src/helpers/listingEvents.ts
```diff
@@ -0,0 +1,10 @@
+import { Dispatch } from "redux";
+import { getTCR } from "./civilInstance";
+import { addListing } from "../actionCreators/listings";
+
+export async function initializeSubscriptions(dispatch: Dispatch): Promise<void> {
+  const tcr = getTCR();
+  await tcr.allApplicationsEver().forEach(async (listing) => {
```
nickreynolds: using allApplicationsEver is just a hack for now, right?

packages/dapp/src/components/Listings.tsx
```diff
@@ -14,208 +13,100 @@ const StyledDiv = styled.div`
   color: black;
 `;
 
-export interface ListingsState {
-  applications: List<ListingWrapper>;
-  whitelistedListings: List<ListingWrapper>;
-  readyToWhitelistListings: List<ListingWrapper>;
-  inChallengeCommitListings: List<ListingWrapper>;
-  inChallengeRevealListings: List<ListingWrapper>;
-  canBeUpdatedListings: List<ListingWrapper>;
```
am9u: Looks like the `canBeUpdatedListings` phase -- which I understand to be 'can Resolve Challenge' got dropped in this refactor. 

Is this no longer a valid phase? Is it wrapped into `awaitingAppealRequestListings`?

walfly: Nothing was ever populating that list, so i killed it

PR #153, opened 2018-05-09
updating core with latest contract changes

nickreynolds:
running migrations with latest contract changes, updating wrappers to support changes


packages/core/src/contracts/tcr/parameterizer.ts
```diff
@@ -274,18 +248,11 @@ export class Parameterizer extends BaseWrapper<ParameterizerContract> {
...
-        return {
-          propID: e.args.propID,
-          paramName: e.args.name,
-          proposedValue: e.args.value,
-          pollID: e.args.pollID,
-        };
-      });
+      ._NewChallengeStream({}, { fromBlock })
+      .concatFilter(async e => this.isChallengeResolved(e.args.challengeID))
+      .map(e => e.args.propID);
```
walfly: whats the thought around changing to just the string?

packages/core/src/contracts/tcr/parameterizer.ts
```diff
@@ -525,7 +492,7 @@ export class Parameterizer extends BaseWrapper<ParameterizerContract> {
    * @param propID ID of prop to check
    */
   public async getPropProcessBy(propID: string): Promise<Date> {
-    const expiryTimestamp = await this.instance.getPropProcessBy.callAsync(propID);
+    const [, , , , , expiryTimestamp] = await this.instance.proposals.callAsync(propID);
```
walfly: gross

packages/core/src/contracts/tcr/parameterizer.ts
```diff
@@ -534,22 +501,25 @@ export class Parameterizer extends BaseWrapper<ParameterizerContract> {
...
+    return name;
   }
 
   /**
    * Returns the value proposed associated with the given proposal
    * @param propID ID of prop to check
    */
   public async getPropValue(propID: string): Promise<BigNumber> {
-    return this.instance.getPropValue.callAsync(propID);
+    const [, , , , , , value] = await this.instance.proposals.callAsync(propID);
```
walfly: dear lord

packages/core/src/contracts/tcr/parameterizer.ts
nickreynolds: when we updated the event to match AdChain standard, we lost a couple of the parameters in the event that help us make the ParamProp, so we just return the ID and get the ParamProp when you need it.

packages/core/src/contracts/tcr/parameterizer.ts
nickreynolds: if you have a better way of doing this, let me know. This is generally how we've been doing it though.

PR #164, opened 2018-05-15
DApp: User can Propose Reparameterization

am9u:
This PR also moves the Parameterizer State into Redux


packages/dapp/src/apis/civilTCR.ts
```diff
@@ -162,3 +174,13 @@ export async function withdrawTokens(address: EthAddress, numTokens: BigNumber):
   const tcr = civil.tcrSingletonTrusted();
   return tcr.withdraw(address, numTokens);
 }
+
+export async function proposeReparameterization(
+  paramName: string,
+  newValue: BigNumber,
```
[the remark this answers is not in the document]
am9u: Good catch! I tested locally with the dev build. The transactions all go through, but I'll try this with the built+served version before I move forward.

packages/dapp/src/components/Parameterizer.tsx
```diff
@@ -1,129 +1,131 @@
...
+          paramKeys={Object.keys(this.props.parameters)}
+          pMinDeposit={this.props.parameters.pMinDeposit && this.props.parameters.pMinDeposit.toString()}
+        />
+      </PageView>
+    );
+  }
 }
 
-export default Parameterizer;
+const mapToStateToProps = (state: State, ownProps: ParameterizerPageProps): object => {
```
walfly: not using ownprops, return type should be specified

packages/dapp/src/helpers/parameterizer.ts
```diff
@@ -0,0 +1,28 @@
+import { Dispatch } from "redux";
+import { setParameter } from "../actionCreators/parameterizer";
+import { getParameterValue } from "../apis/civilTCR";
+
+export async function initializeParameterizer(dispatch: Dispatch): Promise<void> {
```
walfly: how are we keeping the parameter page in sync? is there an event we can subscribe to

PR #166, opened 2018-05-16
Reduxlistingpage

walfly:


packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -108,6 +108,79 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
   }
 
   //#region EventStreams
+  /**
+   * An unending stream of all addresses currently whitelisted
```
nickreynolds: comment wrong

packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -119,8 +192,7 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
     return this.instance
       ._ApplicationWhitelistedStream({}, { fromBlock })
       .map(e => new Listing(this.ethApi, this.instance, e.args.listingAddress))
-      .concatMap(async l => l.getListingWrapper())
-      .concatFilter(l => isWhitelisted(l.data));
+      .concatMap(async l => l.getListingWrapper());
```
nickreynolds: this is no longer returning an observable that only includes currently whitelisted newsrooms..

packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -133,8 +205,7 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
     return this.instance
       ._ApplicationStream({}, { fromBlock })
       .map(e => new Listing(this.ethApi, this.instance, e.args.listingAddress))
-      .concatMap(async l => l.getListingWrapper())
-      .concatFilter(l => isInApplicationPhase(l.data));
+      .concatMap(async l => l.getListingWrapper());
```
nickreynolds: no longer returning observable of newsrooms currently applying..

packages/core/src/contracts/tcr/listing.ts
```diff
@@ -46,76 +46,72 @@ export class Listing {
 
   //#region EventStreams
 
-  public applications(): Observable<TimestampedEvent<CivilTCR.LogEvents._Application>> {
-    return this.tcrInstance._ApplicationStream({ listingAddress: this.listingAddress }, { fromBlock: 0 }).map(e => {
+  public applications(fromBlock: number): Observable<TimestampedEvent<CivilTCR.LogEvents._Application>> {
```
nickreynolds: imo all of these should have a default parameter of 0 (or eventually... the block that we deploy the TCR contract at)

packages/dapp/src/apis/civilTCR.ts
```diff
@@ -116,11 +116,36 @@ export async function getNewsroom(address: EthAddress): Promise<any> {
   return newsroom;
 }
 
-export async function getParameterValue(param: string): Promise<BigNumber> {
+export async function getParameterValue(params: string[]): Promise<BigNumber[]> {
```
am9u: Should we change this method name to `multiGetParameterValue` or `getParameterValues` something like that so it's clear when used in code? Or alternatively, keep `getParmeterValue` as a getter for a single param value and add another method called `multiGetParmeterValue`?

packages/dapp/src/components/Parameterizer.tsx
```diff
@@ -110,19 +110,15 @@ class Parameterizer extends React.Component<{}, ParameterStates> {
...
   private initParameterValues = async (): Promise<any> => {
-    const parameterObj = {};
     const keys = Object.keys(this.state);
-    let i = 0;
 
-    keys.forEach(async parameterKey => {
-      const parameterVal = await getParameterValue(parameterKey);
-      parameterObj[parameterKey] = parameterVal.toString();
-      i++;
+    const parameterVals = await getParameterValue(keys);
```
am9u: So fresh and so clean! 

So I created a reducer for the parameterizer state in another PR, but it's based on the old `initParamValues`. Would love to update my PR with this once yours lands.

walfly:
@nickreynolds going to merge when green

PR #167, opened 2018-05-16
Added revisions support to the smart contract

ritave:


packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -6,64 +6,103 @@ import "../zeppelin-solidity/ECRecovery.sol";
...
+  function getContent(uint contentId) external view returns (bytes32 contentHash, string uri, uint timestamp) {
+    Revision[] storage myContent = content[contentId];
+    require(myContent.length > 0);
+
+    Revision storage myRevision = myContent[myContent.length - 1];
+
+    return (myRevision.contentHash, myRevision.uri, myRevision.timestamp);
+  }
+
+  function getRevision(uint contentId, uint revisionId) external view returns (bytes32 contentHash, string uri, uint timestamp) {
```
nickreynolds: should we include an easy way of getting the number of revisions?

walfly:
Also @ritave, need a revise charter method.  I think it should also have a unique event

packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -3,67 +3,199 @@ pragma solidity ^0.4.19;
...
+  uint public contentCount;
+  /**
+  @notice User readable name of this Newsroom
+  */
   string public name;
 
-  function Newsroom(string newsroomName) ACL() public {
+  function Newsroom(string newsroomName, string charterUri, bytes32 charterHash) ACL() public {
     setName(newsroomName);
+    publishContent(charterUri, charterHash);
```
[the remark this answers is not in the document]
nickreynolds: if they don't want to submit a charter when creating contract, these will just be blank, but at least it'll setup article 0 as the charter article.

nickreynolds:
I'm inclined to agree that we should move the signature stuff out of the smart contract. Seems potentially limiting (what if we invent a new multi-party signature... "verifyRevisionSignature" would become useless)

ritave:
@walfly This is something we talked with Nick, if we force the charter to always be article 0 then in terms of access to data publishRevision is enough. We can just filter existing events by charter id. `reviseCharter` would just duplicate functionality with different naming.

@walfly @nickreynolds I like the idea of signatures in the json, yet signing in the smart-contract serves a different function. It *enforces* the governance model we have decided on. If you want to use it, you have to abide it's logic. While the signature in the json is client-specific, it can be bugged, it can be ignored, etc. Both are useful, just a bit different purposes. The amount of eth transactions is the same either way.

@nickreynolds As for the multi-party signatures, etc - it's a solved problem in cryptography, multisig private keys

I'd say let's leave it for now as is, and if we want - use the JSON signing too.

walfly:
Should we support multiple authors then? 

Also a concern is that we’re adding a lot of surface area to the api. It’s less clear how to use it properly

nickreynolds:
seems like the contract would be simpler if we only had `publishContent` and not `publishSignedContent`.

packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -3,67 +3,199 @@ pragma solidity ^0.4.19;
...
+  /**
+  @notice Updates the existing content with a new revision, the content id stays the same while revision id increases afterwards
+  Requires that the content was first published
+  This function can be only called by ROLE_EDITOR or the owner.
+  Can't be used for signed content.
+
+  @dev Emits `RevisionUpdated` event
+  */
+  function updateRevision(uint contentId, string contentUri, bytes32 contentHash) external requireRole(ROLE_EDITOR) {
+    require(!isSigned(contentId));
```
walfly: editors need to be able to publish unsigned revisions on top of signed ones.  There are a bunch of situations where they may not have access to the original author

walfly: This is the only one that's blocking in my opinion

PR #174, opened 2018-05-21
adding sub-tabs for listings in progress

nickreynolds:
also:
-  fixing bug in core related to can challenge be resolved

### DIFF
--- a/packages/core/src/utils/listingDataHelpers/challengeHelper.ts
+++ b/packages/core/src/utils/listingDataHelpers/challengeHelper.ts
@@ -63,5 +63,5 @@ export function canResolveChallenge(challengeData: ChallengeData): boolean {
  * @param challengeData this ChallengeData to check
  */
 export function doesChallengeHaveAppeal(challengeData: ChallengeData): boolean {
-  return !challengeData.appeal || is0x0Address(challengeData!.appeal!.requester.toString());
+  return challengeData.appeal! && !is0x0Address(challengeData!.appeal!.requester.toString());
 }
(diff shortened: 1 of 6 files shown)


packages/core/src/utils/listingDataHelpers/challengeHelper.ts
```diff
@@ -63,5 +63,5 @@ export function canResolveChallenge(challengeData: ChallengeData): boolean {
  * @param challengeData this ChallengeData to check
  */
 export function doesChallengeHaveAppeal(challengeData: ChallengeData): boolean {
-  return !challengeData.appeal || is0x0Address(challengeData!.appeal!.requester.toString());
+  return challengeData.appeal! && !is0x0Address(challengeData!.appeal!.requester.toString());
```
am9u: Durrrr. Thanks for fixing this.

packages/dapp/src/components/listinglist/ListingsInProgress.tsx
```diff
@@ -0,0 +1,91 @@
...
+  awaitingAppealChallengeListings: Set<string>;
+  appealChallengeCommitPhaseListings: Set<string>;
+  appealChallengeRevealPhaseListings: Set<string>;
+  resolveChallengeListings: Set<string>;
+  resolveAppealListings: Set<string>;
+}
+
+class ListingsInProgress extends React.Component<ListingProps> {
+  public render(): JSX.Element {
+    const applications = this.props.applications;
```
walfly: should these be represented in the reducers?  Like instead of merging them here, get rid of the smaller categories and replace them with meaningful sets?

nickreynolds: I'd rather keep the small categories at least for now. Don't think we're even sure what exactly the bigger categories would be, this makes it a bit easier to change things around?

PR #175, opened 2018-05-21
 DApp: User can view list of Parameterizer proposals

am9u:
This update adds reducers, action generators and views for rendering a list of Parameterizer Proposals on the Parameterizer page of the DApp.


packages/dapp/src/components/parameterizer/proposeReparameterization.tsx
```diff
@@ -64,6 +64,15 @@ export class ProposeReparameterization extends React.Component<
     );
   }
 
+  public componentWillReceiveProps(nextProps: ProposeReparameterizationProps): void {
```
walfly: I'd prefer if you changed this to `getDerivedStateFromProps` now.  The only difference is that you return the object instead of passing it to setState

am9u: Sure. I wasn't sure if the method was available in our current version of React, but I can make this change.

walfly: I guess your right it was added in 16.3.  We should upgrade to that. Its been out for couple months now

am9u: @walfly: is upgrading a blocker for this PR? 

I'd prefer to do an upgrade as a separate PR, since there might be a bunch of changes to get in line with the new React methods.

walfly: its a minor so it will just add warnings, but go ahead and do it separately.

PR #178, opened 2018-05-22
Listing List Item Statuses, States Update Based on Timers

nickreynolds:
- Keeping listings up to date by re-adding them to redux after waiting for expiration timer. 
- Adding statuses to listing list item widget. 
- Fixing bugs in core around getting state.

### DIFF
--- a/packages/core/src/contracts/tcr/civilTCR.ts
+++ b/packages/core/src/contracts/tcr/civilTCR.ts
@@ -22,7 +22,7 @@ import {
   canChallengeBeResolved,
   isAwaitingAppealJudgment,
   isListingAwaitingAppealChallenge,
-  isListingInAppealChallengeCommitPhase,
+  isInAppealChallengeCommitPhase,
   isInAppealChallengeRevealPhase,
   canListingAppealBeResolved,
 } from "../../utils/listingDataHelpers/listingHelper";
@@ -309,7 +309,7 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
       ._ChallengeStream({}, { fromBlock })
       .map(e => new Listing(this.ethApi, this.instance, e.args.listingAddress))
       .concatMap(async l => l.getListingWrapper())
-      .concatFilter(l => isListingInAppealChallengeCommitPhase(l.data));
+      .concatFilter(l => isInAppealChallengeCommitPhase(l.data));
   }
 
   /**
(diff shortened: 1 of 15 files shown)


packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -309,7 +309,7 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
       ._ChallengeStream({}, { fromBlock })
       .map(e => new Listing(this.ethApi, this.instance, e.args.listingAddress))
       .concatMap(async l => l.getListingWrapper())
-      .concatFilter(l => isListingInAppealChallengeCommitPhase(l.data));
+      .concatFilter(l => isInAppealChallengeCommitPhase(l.data));
```
walfly: i think these concat filters are bad because they obscure the number of events actually being fired.

packages/dapp/src/components/listinglist/ListingListItemOwner.tsx
```diff
@@ -26,7 +27,14 @@ class ListingListItemOwner extends React.Component<ListingListItemOwnerProps> {
       <StyledDiv>
         <SectionHeader>OWNER</SectionHeader>
         <br />
-        {owners}
+        {owners.map(owner => {
+          return (
+            <div key={owner + "" + this.props.newsroom.address}>
```
walfly: why not just owner?  It makes no sense to add a consistent value to a key

walfly: honestly i'd just use index

packages/dapp/src/components/listinglist/ListingListItemStatus.tsx
```diff
@@ -18,16 +33,123 @@ class ListingListItemStatus extends React.Component<ListingListItemStatusProps>
...
+    } else if (isListingAwaitingAppealChallenge(listingData)) {
+      return this.renderAwaitingAppealChallenge();
+    } else if (canListingAppealBeResolved(listingData)) {
+      return this.renderAppealCanBeResolved();
+    } else if (isInAppealChallengeCommitPhase(listingData)) {
+      return this.renderAppealChallengeCommitPhase();
+    } else if (isInAppealChallengeRevealPhase(listingData)) {
+      return this.renderAppealChallengeRevealPhase();
+    } else if (canListingAppealChallengeBeResolved(listingData)) {
+      return this.renderAppealChallengeCanBeResolved();
```
walfly: :facemelts:

packages/core/src/contracts/tcr/civilTCR.ts
nickreynolds: well let's talk about that later, not a part of this PR

PR #183, opened 2018-05-22
Current User's Newsrooms

nickreynolds:
- adding set of newsrooms current user is owner of to state
- adding tab to display this list

### DIFF
--- a/packages/dapp/src/helpers/listingEvents.ts
+++ b/packages/dapp/src/helpers/listingEvents.ts
@@ -1,7 +1,7 @@
 import { Dispatch } from "react-redux";
 import { getTCR, getCivil } from "./civilInstance";
 import { addListing } from "../actionCreators/listings";
-import { addNewsroom } from "../actionCreators/newsrooms";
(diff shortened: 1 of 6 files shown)


packages/dapp/src/helpers/listingEvents.ts
```diff
@@ -24,9 +24,13 @@ export async function initializeSubscriptions(dispatch: Dispatch<any>): Promise<
 
 export async function getNewsroom(dispatch: Dispatch<any>, address: EthAddress): Promise<void> {
   const civil = getCivil();
+  const user = civil.userAccount;
   const newsroom = await civil.newsroomAtUntrusted(address);
   const newsroomWrapper = await newsroom.getNewsroomWrapper();
   dispatch(addNewsroom(newsroomWrapper));
+  if (user && newsroomWrapper.data.owners.includes(user)) {
```
am9u: Ah cool. I should prob update the check that displays `Deposit/Withdraw/Exit Listing` on the Listing page for the Newsroom's owners to use this.

PR #184, opened 2018-05-23
Support multisig newsroom TCR application [ch697]

tobek:
- Update abi-gen contract wrapper generators to create multisig wrappers for `EIP20` and `CivilTCR`
    - Existing multisig wrapper for newsroom automatically checked if newsroom owner address was multisig - wrappers can now instead optionally be supplied multisig address when created
- Core `CivilTRC` and `EIP20` now have, like `Newsroom`, an optional `multisigProxy` property that's generated if instantiated with multisig address
- Not every function is run through multisig proxy, right now just token `approve` and TCR `apply`, but more will need to be proxied soon probably


packages/core/src/civil.ts
```diff
@@ -142,6 +142,16 @@ export class Civil {
     return CivilTCR.singleton(this.ethApi, this.contentProvider);
   }
 
+  /**
+   * Same as `initializeMultisig` but supports (but does not require) a multisig proxy.
+   * @param multisigAddress (optional) Multisig through which to proxy interactions with this TCR
+   * @returns Promise containing singleton instance of TCR living on the current network
+   * @throws {CivilErrors.UnsupportedNetwork} In case we're trying to get a non-deployed singleton
+   */
+  public async tcrSingletonTrustedMultisigSupport(multisigAddress?: EthAddress): Promise<CivilTCR> {
```
tobek: Maybe a bit weird creating this function, but the existing `tcrSingletonTrusted` function is synchronous and this one needs to be async, and it seemed silly to update the dozens of existing calls to `tcrSingletonTrusted` (and every branch that leads to those calls) to be async too, especially when only like 2 of them actually need multisig.

packages/core/package.json
```diff
@@ -18,7 +18,7 @@
     "generate:contracts": "abi-gen --abis 'src/artifacts/**/*.json' --template 'src/templates/contract/contract.handlebars' --partials 'src/templates/contract/partials/**/*.handlebars' --output 'src/contracts/generated/wrappers/'",
     "generate:artifacts": "node ../dev-utils/build/scripts/generate-from-files/bin.js 'src/artifacts/**/*.json' 'src/templates/artifacts.handlebars' 'src/contracts/generated/artifacts.ts'",
     "generate:events": "node ../dev-utils/build/scripts/generate-from-files/bin.js 'src/artifacts/**/*.json' 'src/templates/events.handlebars' 'src/contracts/generated/events.ts'",
-    "generate:multisig": "abi-gen --abis 'src/artifacts/Newsroom.json' --template 'src/templates/contract/multisigproxy.handlebars' --partials 'src/templates/contract/partials/**/*.handlebars' --output 'src/contracts/generated/multisig/'",
+    "generate:multisig": "abi-gen --abis 'src/artifacts/@(Newsroom|EIP20|CivilTCR).json' --template 'src/templates/contract/multisigproxy.handlebars' --partials 'src/templates/contract/partials/**/*.handlebars' --output 'src/contracts/generated/multisig/'",
```
ritave: How about doing it this way so it can be reused, easily noticable

```json
{
  "generate:multisig": "... 'src/artifacts/${npm_package_config_multsigGlob}.json' ..."
},
"config: {
  "multisigGlob": "@(Newsroom|EIP20|CivilTCR)"
}

File: packages/core/src/civil.ts [hunk 3]
```diff
@@ -142,6 +142,16 @@ export class Civil {
     return CivilTCR.singleton(this.ethApi, this.contentProvider);
   }
 
+  /**
+   * Same as `initializeMultisig` but supports (but does not require) a multisig proxy.
```
ritave: If the intention is that the default will use multisig, could we use the same naming scheme as in Newsroom? Default with the proxy, explicitly without. If you'd prefer this way, change the newsroom name too.

packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -420,7 +443,8 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
     listingAddress: EthAddress,
     deposit: BigNumber,
     applicationContent: string,
-  ): Promise<TwoStepEthTransaction> {
+  ): Promise<MultisigProxyTransaction> {
+    // TODO(tobek) Confirm that multisig proxy isn't needed for generating this URI
```
ritave: I confirm that it's not needed

packages/core/src/contracts/multisig/basemultisigproxy.ts
```diff
@@ -8,29 +9,40 @@ import { artifacts } from "../generated/artifacts";
...
 export class BaseMultisigProxy {
   protected ethApi: EthApi;
+  protected contractOwner?: EthAddress;
   protected multisig?: Multisig;
   // TODO(ritave): Add support for lowercase newsroom contract in abi-gen
-  protected instance: OwnableContract;
+  protected instance: Contract | OwnableContract;
 
-  protected constructor(ethApi: EthApi, instance: OwnableContract) {
+  protected constructor(ethApi: EthApi, instance: Contract | OwnableContract, contractOwner?: EthAddress) {
```
ritave: This contract doesn't need to touch anything concerning Ownable now. `OwnableContract` can be removed altogether

packages/core/src/contracts/multisig/basemultisigproxy.ts
```diff
@@ -40,12 +52,24 @@ export class BaseMultisigProxy {
     return undefined;
   }
 
-  protected async resolveMultisig(): Promise<void> {
-    const owner = await this.instance.owner.callAsync();
+  /**
+   * Instantiate `this.multisig` if appropriate.
+   * @param multisigAddress (optional) If supplied, instantiate multisig from that, otherwise check if `this.contractOwner` is a multisig wallet and instantiate from that. Otherwise `this.multisig` remains undefined.
```
ritave: Shouldn't `multisigAddress == contractOwner` in all situations? I think we can simplify this to just having `contractOwner` and verifying code there.

Otherwise we have a different owner from the multisig that we control, which is very unexpected logic

packages/core/src/contracts/tcr/eip20.ts
```diff
@@ -30,7 +42,10 @@ export class EIP20 extends BaseWrapper<EIP20Contract> {
    * @param spender address to approve as spender of tokens
    * @param numTokens number of tokens to approve for spender to spend on user's behalf
    */
-  public async approveSpender(spender: EthAddress, numTokens: BigNumber): Promise<TwoStepEthTransaction> {
+  public async approveSpender(spender: EthAddress, numTokens: BigNumber): Promise<MultisigProxyTransaction> {
```
ritave: The user of this library will expect all mutations to work the same way. Might as well give `transfer` the multisig treatment too

packages/core/src/contracts/tcr/eip20.ts
```diff
@@ -52,7 +68,7 @@ export class EIP20 extends BaseWrapper<EIP20Contract> {
 
   /**
    * Check the token balance of an address
-   * @param address address to check balance of
+   * @param tokenOwner address to check balance of (defaults to current user)
```
ritave: Should it default to current user or the multisig address.

packages/core/src/templates/contract/multisigproxy.handlebars
```diff
@@ -30,8 +46,12 @@ export class {{contractName}}MultisigProxy extends BaseMultisigProxy {
   {{/this.constant}}
 {{/each}}
 
-  private constructor(ethApi: EthApi, instance: OwnableContract) {
-    super(ethApi, instance);
-    bindNestedAll(this, ["constructor", "instance", "multisig", "ethApi"]);
+  private constructor(
+    ethApi: EthApi,
+    instance: Contract | OwnableContract,
```
ritave: just `Contract` is enough

packages/core/package.json
walfly: that works?

ritave: @walfly Check out root package.json's ganache script

packages/core/src/contracts/multisig/basemultisigproxy.ts
tobek: This class used to make several calls to `this.instance.owner()` (defined on `OwnableContract` but not `Contract`), but I moved all that logic into the multisigproxy wrapper `create` function, see line `contractOwner = await (instance as OwnableContract).owner.callAsync();` and how `contractOwner` is used here. I think there's a better refactor but as it stands we can't totally do away with `OwnableContract` yet.

packages/core/src/contracts/multisig/basemultisigproxy.ts
tobek: You're right, I did it this way to support the previous logic, which supported *non* multisig newsrooms, where `contractOwner` would not be a multisig address, but just some random wallet.

It's maybe also confusing because in the case of e.g. `EIP20MultsigProxy` there isn't really a `contractOwner` - that's only relevant for `NewsroomMultisigProxy`.

Can we completely stop supporting non-multisig newsrooms? If so that makes cleaning this up easier.

packages/core/src/contracts/tcr/eip20.ts
tobek: You're right, will update, though the dapp currently uses `transfer` for transferring from logged-in user *to* the multisig so we still need to support non-multisig. I'm going to update comments and variable names to clearly distinguish between `tcrViaMultisig`/`tokenViaMultisig` vs. non-multisig-instantiated ones, and library user needs to keep track of which kind they've created.

tobek: Actually will update all of the functions in here, including `getBalance` etc., to default to the multisig address if this is a multisig instance.

packages/core/src/contracts/tcr/eip20.ts
tobek: Multisig, if multisig instance - will update.

packages/core/src/civil.ts
tobek: (Typo in my comment, should have read "Same as `tcrSingletonTrusted`, but supports...")

The issue is that the vast majority of instantiations of the TCR do not require multisig - most of the stuff in TCR API here is read-only. Additionally, `tcrSingletonTrusted` is synchronous, but this function has to be async, and there are 100+ synchronous pathways that lead to this function (e.g. every `getCivil()` call in dapp) , and updating them all to be async all the way through chain of calls seemed crazy.

I've actually created a synchronous `createNonMultisig` function on multisig wrapper so that we can always rely on a multisig proxy being present with the same API, even if it's not actually using multisig. On the one hand this feels weird, on the other hand it prevents dozens of `if (this.multisigProxy) { ... } else { .... }`.

packages/core/src/templates/contract/multisigproxy.handlebars
```diff
@@ -2,25 +2,49 @@
...
+    await proxy.resolveMultisig(multisigAddress);
+    return proxy;
+  }
+
+  /**
+   * Creates a wrapper around a contract with identical API to multisig wrapper, but all transactions are passed directly to contract. Essentially just a helper for when we know multisig isn't needed and want a synchronous call.
+   * @param ethApi Web3 API
+   * @param instance Contract we are wrapping
+   */
+  public static createNonMultisig(
```
tobek: Just copy-pasting my comment about this since the comment thread is hidden now because the code it was on is gone.

The reason I made function is to allow a synchronous way of instantiating TCR/token when multisig is not needed, which is 95% of the time. Existing `tcrSingletonTrusted` is synchronous but multisig version must be async, and there are 100+ synchronous pathways that lead to `tcrSingletonTrusted` (e.g. every `getCivil()` call in dapp) , and updating them all to be async all the way through chain of calls seemed crazy.

We need this also so we can always rely on a multisig proxy being present with the same API, even if it's not actually using multisig. On the one hand this feels weird, on the other hand it prevents dozens of `if (this.multisigProxy) { ... } else { .... }` branches.

tobek:
Update: put non-multisig support back in `core`, wasn't hard in the end, only fudge needed to make this happen was to add optional `owner` property to `Contract` interface. This way both 1) a newsroom contract can be sent to multisig proxy whether it's multisig or not and the proxy can figure it out if `owner` is present, and 2) non-ownable contracts like `CivilTCR` can be sent to multisig proxy with or without `owner` and without multisig enabled.

This PR should be ready to merge unless there's any other feedback.

ritave:
@tobek Support for non-multisig Newsrooms is quite useful debugging facility, we wanted to remove it but was left on purpose

packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -532,19 +558,19 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
...
   }
 
-  public async challengeGrantedAppeal(listingAddres: EthAddress, data: string = ""): Promise<TwoStepEthTransaction> {
-    return createTwoStepSimple(
-      this.ethApi,
-      await this.instance.challengeGrantedAppeal.sendTransactionAsync(listingAddres, data),
-    );
+  public async challengeGrantedAppeal(
+    listingAddress: EthAddress,
+    data: string = "",
```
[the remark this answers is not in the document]
tobek: This line is only changed because the return type changed to longer `MultisigProxyTransaction` so prettier broke params onto separate lines. Not sure what this will be in the end so I'll leave it for now.

packages/core/src/contracts/interfaces/contract.ts
```diff
@@ -2,4 +2,5 @@ import { EthAddress } from "../../types";
 
 export interface Contract {
   readonly address: EthAddress;
+  owner?: { callAsync(): Promise<EthAddress> };
```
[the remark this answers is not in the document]
tobek: Hell yeah this is exactly what I was looking for but my typescript was rusty.

That particular one didn't work (`Type 'OwnableContract' is not assignable to type 'T'` which is... not true) but casting it (which typescript docs recommends for type guard predicates) works:

```typescript
public isOwnable(contract: Contract | OwnableContract): contract is OwnableContract {
  return (contract as OwnableContract).owner !== undefined;
}
```

Anyways I'm putting it in `utils/contracts`.

PR #186, opened 2018-05-24
Hotfixes for bugs in the Reparameterization Proposal view

am9u:
This updates fixes a couple of bugs that were preventing the Reparameterization Proposal view from rendering, which in turn would prevent users acting on proposals.

### DIFF
--- a/packages/dapp/src/components/parameterizer/Proposal.tsx
+++ b/packages/dapp/src/components/parameterizer/Proposal.tsx
@@ -27,7 +27,6 @@ export interface ProposalReduxProps {
 
 class Proposal extends React.Component<ProposalPageProps & ProposalReduxProps> {
   public render(): JSX.Element {
-    console.log(this.props.proposal);
     return (
       <PageView>
         <ViewModule>
--- a/packages/dapp/src/components/parameterizer/Proposal.tsx
+++ b/packages/dapp/src/components/parameterizer/Proposal.tsx
@@ -27,6 +27,7 @@ export interface ProposalReduxProps {
 
 class Proposal extends React.Component<ProposalPageProps & ProposalReduxProps> {
   public render(): JSX.Element {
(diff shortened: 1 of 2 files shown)


packages/dapp/src/components/parameterizer/Proposal.tsx
```diff
@@ -90,7 +92,8 @@ class Proposal extends React.Component<ProposalPageProps & ProposalReduxProps> {
   private renderUpdateParam = (): JSX.Element => {
     return (
       <StyledFormContainer>
-        Parameter Update Phase ends in <CountdownTimer endTime={this.props.proposal.propProcessByExpiry.toNumber()} />
+        Parameter Update Phase ends in{" "}
+        <CountdownTimer endTime={this.props.proposal.propProcessByExpiry.valueOf / 1000} />
```
nickreynolds: valueOf missing () here

packages/dapp/src/helpers/parameterizer.ts
```diff
@@ -48,8 +48,12 @@ export async function initializeProposalsSubscriptions(dispatch: Dispatch<any>):
     const propState = await parameterizer.getPropState(propID);
     const challengeID = await parameterizer.getChallengeID(propID);
     const applicationExpiry = await parameterizer.getPropApplicationExpiry(propID);
-    const challengeCommitExpiry = await parameterizer.getPropChallengeCommitExpiry(propID);
-    const challengeRevealExpiry = await parameterizer.getPropChallengeRevealExpiry(propID);
+    const challengeCommitExpiry = !challengeID.isZero()
+      ? await parameterizer.getPropChallengeCommitExpiry(propID)
+      : null;
```
nickreynolds: i think you're supposed to use `undefined` instead of `null` pretty much always in typescript land, not sure it really matters though

PR #190, opened 2018-05-30
user switching

nickreynolds:
- Adding polling for web3 account. 
- When account changes, get new balances and challenge data. 
- Updating redux stores to support data for different users.

### DIFF
--- a/packages/core/src/utils/ethapi.ts
+++ b/packages/core/src/utils/ethapi.ts
@@ -43,11 +43,9 @@ export class EthApi {
     //               It shouldn't, and should just set null account
     this.currentProvider = provider;
     this.abiDecoder = new AbiDecoder(Object.values<Artifact>(artifacts).map(a => a.abi));
-    if (onAccountSet) {
-      this.onAccountSet = onAccountSet;
-      setInterval(this.accountPing, 100);
-      this.accountPing();
-    }
+    this.onAccountSet = onAccountSet;
+    setInterval(this.accountPing, 100);
+    this.accountPing();
   }
 
   public get currentProvider(): Web3.Provider {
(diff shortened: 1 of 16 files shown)


packages/core/src/utils/ethapi.ts
```diff
@@ -226,4 +234,14 @@ export class EthApi {
       throw new Error(CivilErrors.EvmException);
     }
   }
+
+  private setAccount = (newAccount: EthAddress): void => {
```
walfly: To make this a little more generic you could use the events package.

```
class EthApi extends Events {
  public set account(): void {
      // ...whatever
      this.trigger("set:account", data);
  }
}
```
then you could bind to it at any time instead of passing it in at construction only.  Probably want to put the actual event on Civil.ts

packages/dapp/src/components/Main.tsx
```diff
@@ -20,17 +20,24 @@ import { connect, DispatchProp } from "react-redux";
...
     await initializeParameterizer(this.props.dispatch!);
     await initializeGovernment(this.props.dispatch!);
     await initializeProposalsSubscriptions(this.props.dispatch!);
     await initializeGovernmentParamSubscription(this.props.dispatch!);
     await initializeSubscriptions(this.props.dispatch!);
+  }
+
+  public onAccountUpdated = async (): Promise<void> => {
+    const civil = getCivil();
+    const tcr = civil.tcrSingletonTrusted();
```
am9u: This is more for my own edification, but sometimes I see this in the code:

```
const civil = getCivil();
const tcr = civil.tcrSingletonTrusted();
```

And sometimes I see this:

```
const tcr = getTCR();
```

Is one preferable to the other? Or does the method of getting `tcr` differ in different situations?

nickreynolds: no real difference, the 2nd one is just shorthand. I think we need to rethink slightly how we handle the civil instance, so I'm not too concerned about this since I think it's going to change anyways.

PR #191, opened 2018-05-31
Create Newsroom: User can sign the Constitution

am9u:
This update adds the ability for a Newsmaker to sign the Constitution, assuming they're an owner of the Newsroom.

There are a few things in this update that I had questions on / could use an eye, so please read the inline comments in the PR for details.


packages/core/src/civil.ts
```diff
@@ -60,6 +61,10 @@ export class Civil {
     return this.ethApi.toBigNumber(num);
   }
 
+  public async signMessage(message: string): Promise<EthSignedMessage> {
+    return this.ethApi.signMessage(message, this.userAccount);
```
am9u: Is this the correct method to use for signing the Constitution?

@ritave, I think you're the knowledgable party here

packages/core/src/civil.ts
```diff
@@ -60,6 +61,10 @@ export class Civil {
     return this.ethApi.toBigNumber(num);
   }
 
+  public async signMessage(message: string, account?: EthAddress): Promise<EthSignedMessage> {
+    return this.ethApi.signMessage(message, account);
```
am9u: Is this the correct method to use for signing the Constitution?

@ritave, I think you're the knowledgable party here

packages/core/src/utils/ethapi.ts
```diff
@@ -128,7 +128,8 @@ export class EthApi {
 
     const signerAccount = account || requireAccount(this);
 
-    const response = await this.rpc("eth_sign", [signerAccount, messageHex]);
+    // const response = await this.rpc("eth_sign", [signerAccount, messageHex]);
+    const response = await this.rpc("personal_sign", messageHex, signerAccount);
```
am9u: Two things here: 

1) The commented-out line was causing an error in MetaMask/Web3's `eth_sign` stack. That code was expecting 2 separate string args, but the commented out line is passing 1 array arg (that has 2 string members). So the first change I made was to change the args that are being passed. Cool?

2) @tobek mentioned that `web3.personal.sign` is better to use for signing, so I made this change. Is that cool / compatible? Or are there any things upstream/downstream that this will break?

@ritave, I think you're also the knowledgeable party here...

packages/dapp/src/reducers/ui.ts
```diff
@@ -0,0 +1,12 @@
+import { AnyAction } from "redux";
+import { uiActions } from "../actionCreators/ui";
+import { Map } from "immutable";
+
+export function ui(state: Map<string, any> = Map<string, any>(), action: AnyAction): Map<string, any> {
```
am9u: I added action/reducer for storing UI state in Redux, which I think we'll need at some point.

I didn't go down the path of scoping out what the UI state structured data should look like, so right now it's a simple Map that holds key/vals that the multiple View containers/components would be interested in.

@walker Thoughts on this?

We could enhance this by coming up with some sort of structure for the UI state data... like maybe `ui` breaks down into something like:

```
ui.pages = Map<string, Map<string, any>]
// Top level map uses the Component referenced by `Router` as the key.
// The internal maps are key/val pairs relevant to the Component. We could additionally have a 'global' namespace for key/vals that are shared throughout the entire dapp

Or at the very least, keep it simple like in this PR but create an enum with the keys for UI state values.

File: packages/dapp/src/components/newsroom/NameAndAddress.tsx [hunk 5]
```diff
@@ -42,13 +45,22 @@ export class NameAndAddress extends React.Component<StepProps, NameAndAddressSta
...
-  //   this.props.history.push("/mgmt/" + address);
-  // };
+  private onNewsroomCreated = async (result: any) => {
+    const civil = getCivil();
+    const user = civil.userAccount;
+    const newsroomWrapper = await result.getNewsroomWrapper();
+    this.props.dispatch!(addNewsroom(newsroomWrapper));
+    if (user && newsroomWrapper.data.owners.includes(user)) {
+      this.props.dispatch!(addUserNewsroom(result.address));
+      this.props.dispatch!(addOrUpdateUIState("newsroomMgmtCurrentAddress", result.address));
```
am9u: Please see my comments in `src/reducers/ui.ts`!

packages/dapp/src/components/newsroom/SignConstitution.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  public async componentDidMount(): Promise<void> {
+    return this.initGovernmentData();
+  }
+
+  public async componentDidUpdate(
+    prevProps: StepProps & SignConstitutionReduxProps,
+    prevState: SignConstitutionState,
+  ): Promise<void> {
+    const prevNewsroomAddress = prevProps.ui.get("newsroomMgmtCurrentAddress");
```
am9u: Please see my comments in `src/reducers/ui.ts`!

packages/dapp/src/components/newsroom/SignConstitution.tsx
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  private signConstitution = async (): Promise<void> => {
+    const signature = await signMessage(this.props.government.get("constitutionHash"));
+    console.log(signature);
+  };
+}
+
+const mapStateToProps = (state: State, ownProps: StepProps): SignConstitutionReduxProps => {
+  const { ui, government, user } = state;
```
am9u: Please see my comments in `src/reducers/ui.ts`!

packages/dapp/src/components/newsroom/NameAndAddress.tsx
walfly: dang, I did this too but in a different way

packages/core/src/civil.ts
ritave: Have we defined what "singing the Constitution" really means?
This function is correct in sense that it uses the currently open wallet to sign a message you provide with a cryptographic proof that it was this wallet.

But then who is the wallet in the first place? Is it one of the owners? Should they all sign it?

For now I'd say this is the best way until (if) we want to make a bigger ceremony out of signing the constitution

packages/core/src/utils/ethapi.ts
ritave: Right now signing is bit of a mess.
`personal_sign` only works for utf8 strings, which is not the case when for example we're approving articles on newsrooms.

I think for now the library should support both as two API endpoints and use one that works for your use-case

See:
https://github.com/0xProject/0x-monorepo/issues/162
https://github.com/ethereum/EIPs/pull/683
https://medium.com/metamask/the-new-secure-way-to-sign-data-in-your-browser-6af9dd2a1527

packages/components/src/SignConstitutionButton.stories.tsx
```diff
@@ -0,0 +1,96 @@
+import { storiesOf } from "@storybook/react";
+import * as React from "react";
+import styled from "styled-components";
+import { SignConstitutionButton } from "./SignConstitutionButton";
+import { Civil } from "@joincivil/core";
+
+let civil: Civil | undefined;
+
+try {
+  civil = new Civil();
```
ritave: Wasn't there a civil provider pattern somewhere in dapp?

packages/components/src/SignConstitutionButton.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+  max-width: 500px;
+`;
+
+storiesOf("Sign Constitution Button", module)
+  .add("User can sign", () => {
+    return (
+      <Wrapper>
+        <SignConstitutionButton
+          civil={civil}
+          requiredNetwork="rinkeby"
```
ritave: I'm not sure but the network returned might a number (4 for rinkeby) not the name

packages/dapp/src/reducers/ui.ts
walfly: I think we'll definitely need a ui reducer in the dapp

packages/components/src/SignConstitutionButton.stories.tsx
walfly: this isn't dapp, its patterns

packages/components/src/SignConstitutionButton.stories.tsx
walfly: I added a method to get it by name

walfly:
@am9u I pushed what i have locally, i'm passing in address from newsroom

PR #192, opened 2018-06-01
Couple multisig improvements

tobek:
- Missing `await` made it seem like everything was multisig
- Quick check that there's enough token to apply

### DIFF
--- a/packages/dapp/src/apis/civilTCR.ts
+++ b/packages/dapp/src/apis/civilTCR.ts
@@ -1,6 +1,6 @@
 import BigNumber from "bignumber.js";
 import { getCivil, getTCR } from "../helpers/civilInstance";
-import { TwoStepEthTransaction, EthAddress } from "@joincivil/core";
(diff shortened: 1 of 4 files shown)


packages/dapp/src/apis/civilTCR.ts
```diff
@@ -53,6 +53,9 @@ export async function approve(
   const token = await tcr.getToken();
   const amountBN = ensureWeb3BigNumber(amount);
   console.log("token address: " + token.address);
+  if ((await token.getBalance()) < amountBN) {
```
nickreynolds: probably worth adding the CVL -> multisig transfer into the series of apply transactions. not a big deal though.

tobek: Agreed, or we should at least offer to do it - and then check that you have enough CVL. But potentially the person applying might not be the person with CVL to send to multisig, so might have to happen separately. We can work out that flow later.

PR #206, opened 2018-06-08
Manager page is its own package 

walfly:


packages/newsroom-manager/src/SignConstitution.tsx
```diff
@@ -0,0 +1,145 @@
...
+    // }
+    // if (Object.keys(newState).length) {
+    // this.setState(newState);
+    // }
+  }
+
+  public render(): JSX.Element {
+    // const civil = getCivil();
+    return (
+      <StepStyled disabled={this.props.disabled} index={this.props.index || 0}>
```
am9u: Just to verify, the "coming soon" steps haven't been converted to use `CivilContext` and work within the Newsroom package structure b/c they're coming soon?

walfly: Yeah, I didn't want to move all the civil api stuff in here last night.  So i commented it out to come back to it after launch

PR #208, opened 2018-06-10
Deployed the smart-contracts for rinkeby beta

ritave:
Also tagged the version as protocol@v0-rinkeby.

No further deploys of (Newsroom) contracts should happen before we're ready to change the protocol.

When merged, a new core release should happen and the major version be frozen until we're ready to change protocol


nickreynolds:
can we actually remove the changes to the tcr contracts? doesn't seem like any reason to update them

ritave:
Doesn't seem necessary anymore

PR #210, opened 2018-06-12
Add and implement Listing Card view for the Registry page

am9u:
image

Still some more work to do on the Registry page, but the cards are is prob good to merge and have people start looking at.


packages/components/src/ListingSummary.tsx
```diff
@@ -0,0 +1,160 @@
...
+  display: block;
+  font: normal 14px/17px ${fonts.SANS_SERIF};
+  max-width: 65%;
+  overflow-y: hidden;
+  text-decoration: none;
+  text-overflow: ellipsis;
+`;
+
+export interface ListingSummaryComponentProps {
+  address?: EthAddress;
```
nickreynolds: do you think it's better to pass all these props in individually instead of just pulling them out from a ListingWrapper / Newsroomwrapper in render()?

packages/components/src/ListingSummary.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    if (expiry) {
+      return <CountdownTimer endTime={expiry!} warn={warn} />;
+    }
+
+    return;
+  };
+}
+
+export interface ListingSummaryListComponentProps {
```
nickreynolds: even though it's really small, I'd probably prefer if the list component is separated into its own file. also the `any[]` for the prop type is pretty undescriptive..

packages/components/src/PhaseCountdown.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+const StyledDiv = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  width: 400px;
+`;
+
+const Container: React.StatelessComponent = ({ children }) => <StyledDiv>{children}</StyledDiv>;
+const now = Date.now() / 1000;
+const eightDaysFromNow = now + 86400 * 8;
```
nickreynolds: not super important since this is just a story, but probably worth adding some time-manipulation functions in `utils` package for this type of thing

packages/components/src/PhaseCountdown.tsx
```diff
@@ -0,0 +1,128 @@
...
+    );
+  }
+
+  private renderExpiry = (): JSX.Element => {
+    const expiryDateTime = new Date(this.props.endTime * 1000);
+    const timeString = `${pad(expiryDateTime.getHours(), 2)}:${pad(expiryDateTime.getMinutes(), 2)} GMT-${pad(
+      expiryDateTime.getTimezoneOffset() / 60,
+      2,
+    )}${pad(expiryDateTime.getTimezoneOffset() % 60, 2)}`;
+    // const timeString = expiryDateTime.toTimeString();
```
nickreynolds: left in commented out line

packages/components/src/PhaseCountdown.tsx
```diff
@@ -0,0 +1,128 @@
...
+        {label}
+        {this.renderReadableTimeRemaining()}
+        {this.renderExpiry()}
+      </StyledCountdownTimerContainer>
+    );
+  }
+
+  private renderExpiry = (): JSX.Element => {
+    const expiryDateTime = new Date(this.props.endTime * 1000);
+    const timeString = `${pad(expiryDateTime.getHours(), 2)}:${pad(expiryDateTime.getMinutes(), 2)} GMT-${pad(
```
nickreynolds: most of this function seems like it could be in `utils` too

packages/components/src/PhaseCountdown.tsx
```diff
@@ -0,0 +1,128 @@
...
+      return (
+        <StyledCountdownLabelWarn>
+          <b>{getReadableDuration(this.state.secondsRemaining)}</b>
+        </StyledCountdownLabelWarn>
+      );
+    }
+    return <b>{getReadableDuration(this.state.secondsRemaining)}</b>;
+  };
+
+  // TODO(nickreynolds): move this all into redux
```
nickreynolds: I think this is TO-DID now, yes?

packages/dapp/src/components/listinglist/ListingList.tsx
```diff
@@ -1,20 +1,8 @@
 import * as React from "react";
-import styled from "styled-components";
+// import styled from "styled-components";
```
nickreynolds: remove commented out line

packages/dapp/src/components/listinglist/ListingListItem.tsx
```diff
@@ -1,25 +1,26 @@
 import * as React from "react";
 import { connect } from "react-redux";
-import styled from "styled-components";
+// import styled from "styled-components";
```
nickreynolds: remove commented out lines

packages/dapp/src/components/listinglist/ListingListItem.tsx
```diff
@@ -41,18 +42,39 @@ class ListingListItem extends React.Component<ListingListItemOwnProps & ListingL
...
+      isInApplication,
+      canBeChallenged,
+      inChallengePhase,
+      inRevealPhase,
+      appExpiry,
+      commitEndDate,
+      revealEndDate,
+    };
+
+    // console.log(this.props, listingViewProps);
```
nickreynolds: remove

packages/components/src/ListingSummary.tsx
am9u: I went back and forth about this, but I think this props structure for these very view-specific components would be good for the future.

I think we could eventually have the design team manage these components / styling / etc, and having the components ignorant of the `ListingWrapper` / `NewsroomWrapper` makes them easier to work on in Storybook and apply designs to.

I think we ideally have either a selector layer added to our Redux setup that creates these flat structures for the dumb components... or a shared method that does the translation.

PR #212, opened 2018-06-13
add name change, support usernames if get function supplied

walfly:


packages/newsroom-manager/src/CompleteYourProfile.tsx
```diff
@@ -17,15 +17,16 @@ import { connect, DispatchProp } from "react-redux";
...
+import { NewsroomUser } from "./NewsroomUser";
 
 export interface CompleteYourProfileComponentExternalProps extends StepProps {
   address?: EthAddress;
 }
 
 export interface CompleteYourProfileComponentProps extends StepProps {
-  owners: EthAddress[];
-  editors: EthAddress[];
+  owners: Array<{ address: EthAddress; name?: string }>;
```
am9u: Do we need to start thinking about other structured data fields for Newsroom users, ie Bio, Twitter handle, whatever else?

Or is that stuff going to be addressed as part of The Charter conversation?

FWIW, the designs have UI elements for Newsroom User data points like the one's mentioned above.

https://consensys.invisionapp.com/d/main#/console/14237549/300540757/preview

walfly: :punting:

PR #214, opened 2018-06-15
Network Check

nickreynolds:
- Continously poll for network changes and subscribe to changes. 
- When network changes, clear almost all of redux store and restart subscriptions. 
- Add error bar when not on rinkeby.

### DIFF
--- a/packages/core/src/utils/ethapi.ts
+++ b/packages/core/src/utils/ethapi.ts
@@ -76,7 +76,7 @@ export class EthApi extends Events {
 
   public networkPing = (): void => {
     this.setNetwork(this.web3.version.network);
-  }
+  };
 
   public get networkId(): string {
     return this.currentNetwork;
@@ -85,6 +85,12 @@ export class EthApi extends Events {
   public cancelAccountPing(): void {
     if (interval) {
       clearInterval(interval);
+    }
+  }
+
+  public cancelNetworkPing(): void {
+    if (networkInterval) {
+      clearInterval(networkInterval);
     }
   }
 
@@ -264,5 +270,5 @@ export class EthApi extends Events {
       this.currentNetwork = newNetwork;
       this.emit("networkSet");
     }
-  }
+  };
 }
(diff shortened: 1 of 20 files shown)


packages/core/src/utils/ethapi.ts
```diff
@@ -250,4 +264,11 @@ export class EthApi extends Events {
       this.emit("accountSet");
     }
   };
+
+  private setNetwork = (newNetwork: string): void => {
+    if (newNetwork !== this.currentNetwork) {
+      this.currentNetwork = newNetwork;
+      this.emit("networkSet");
```
walfly: should this emit the network id? `this.emit("networkSet", {networkId: newNetwork})`

packages/dapp/src/reducers/index.ts
```diff
@@ -36,13 +36,21 @@ import {
...
 import { ui } from "./ui";
 import { Set, List, Map } from "immutable";
 import { TimestampedEvent, WrappedChallengeData, UserChallengeData, EthAddress } from "@joincivil/core";
 import { currentUserNewsrooms } from "./newsrooms";
 import { newsrooms, NewsroomState, newsroomUi, newsroomUsers } from "@joincivil/newsroom-manager";
+import { networkActions } from "../actionCreators/network";
 
 export interface State {
+  networkDependent: NetworkDependentState;
+  network: string;
```
walfly: I'd probably put network in the ui reducer but either work

packages/dapp/src/reducers/index.ts
```diff
@@ -117,5 +124,17 @@ export default combineReducers({
   challengeUserData,
   appealChallengeUserData,
   government,
+});
+
+const networkDependent = (state: any, action: AnyAction) => {
+  if (action.type === networkActions.SET_NETWORK) {
+    return networkDependentReducers(undefined, action);
```
walfly: why does this need undefined passed in

nickreynolds: that's what the example I looked up said to do to reset state

packages/core/src/utils/ethapi.ts
nickreynolds: It doesn't really matter, you have immediate access to it anyways. I left this out for now but if we want to add it later should be fine.

PR #215, opened 2018-06-19
sign and recover personal messages

walfly:


walfly:
@nickreynolds @ritave updated

PR #218, opened 2018-06-19
Add Progress Bar Countdown Timer component for Listing Detail pages

am9u:
- Refactor `PhaseCountdown` into `TextCountdownTimer` and a higher-order
component that encapsulates the countdown state logic
- Build out `ProgressBarCountdownTimer` component


packages/components/src/PhaseCountdown/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { ProgressBarCountdownTimerComponent } from "./ProgressBarCountdownTimer";
+import * as React from "react";
+
+const connectCountdownTimer = () => <TCountdownTimerProps extends CountdownTimerProps>(
+  CountdownTimerComponent:
+    | React.ComponentClass<TCountdownTimerProps & InjectedCountdownTimerProps>
+    | React.StatelessComponent<TCountdownTimerProps & InjectedCountdownTimerProps>,
+) => {
+  return class HOCountdownTimerContainer extends React.Component<TCountdownTimerProps, CountdownTimerState> {
+    public _timer: number | undefined;
```
nickreynolds: any reason why this is prefixed with an underscore? also it could probably be `public timer?: number` right?

PR #221, opened 2018-06-20
Button / TransactionButton component style updates

am9u:
- Add DarkButton
- Add Inverted and Dark TransactionButton styles
- Add support for custom pre/post transtions AOP methods to
TransactionButton

### DIFF
--- a/packages/components/src/Button.tsx
+++ b/packages/components/src/Button.tsx
@@ -101,7 +101,7 @@ export const SecondaryButton = BaseButton.extend`
   background-color: ${colors.basic.WHITE};
   color: ${colors.accent.CIVIL_GRAY_2};
   border: 1px solid ${colors.accent.CIVIL_GRAY_3};
-  &.active {
+  &:hover {
     background-color: ${colors.accent.CIVIL_BLUE};
     border: 1px solid ${colors.accent.CIVIL_BLUE};
     color: ${colors.basic.WHITE};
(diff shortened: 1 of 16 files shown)


packages/components/src/TransactionButton.tsx
```diff
@@ -72,6 +73,37 @@ export class TransactionButtonNoModal extends React.Component<TransactionButtonP
   }
 
   public render(): JSX.Element {
+    if (this.props.style && this.props.style === "inverted") {
```
walfly: should this be a switch? or assign a variable like `Button` to the button component you want.

```
let Button: StyledComponentClass;
if (whatever condition) {
  Button = DarkButton;
}
return <Buton {...props}/>;
```

packages/components/src/TransactionButton.tsx
```diff
@@ -72,14 +73,29 @@ export class TransactionButtonNoModal extends React.Component<TransactionButtonP
   }
 
   public render(): JSX.Element {
+    let StyledButton;
```
am9u: Ok, I ended up taking both of your suggestions @walfly.  Thx, it def looks more readable.

packages/components/src/Button.tsx
```diff
@@ -113,6 +113,22 @@ export const SecondaryButton = BaseButton.extend`
...
+  &.active {
+    background-color: ${colors.accent.CIVIL_GRAY_1};
+    border: 1px solid ${colors.accent.CIVIL_GRAY_1};
+    color: ${colors.basic.WHITE};
+  }
+  &.active {
+    background-color: ${colors.accent.CIVIL_GRAY_1};
+    border: 1px solid ${colors.accent.CIVIL_GRAY_1};
+    color: ${colors.basic.WHITE};
+  }
```
sruddy: Should the second `&.active` be a different state? I see it happens on `SecondaryButton` also.

am9u: Ah, good catch. I was copy-pasta-ing those definitions from `SecondaryButton`, but my guess is that they're supposed to be `:hover` definitions in there. I'm updating with that assumption, and we can address further if the design team has feedback.

PR #223, opened 2018-06-20
Transaction Button Fix

nickreynolds:
- reset state of transaction button if transaction has error of some kind
- https://app.clubhouse.io/civil/story/1298/dapp-transaction-button-gets-into-bad-state-when-transaction-fails-or-is-canceled
- https://app.clubhouse.io/civil/story/1299/transaction-button-releases-when-you-reject

### DIFF
--- a/packages/components/src/TransactionButton.tsx
+++ b/packages/components/src/TransactionButton.tsx
@@ -129,6 +129,7 @@ export class TransactionButtonNoModal extends React.Component<TransactionButtonP
         return this.executeTransactions(transactions);
       } catch (err) {
         if (currTransaction.handleTransactionError) {
+          this.setState({ step: 0, disableButton: false });
           setImmediate(() => currTransaction.handleTransactionError!(err));
         }
       }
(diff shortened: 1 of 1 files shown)


packages/components/src/TransactionButton.tsx
```diff
@@ -129,6 +129,7 @@ export class TransactionButtonNoModal extends React.Component<TransactionButtonP
         return this.executeTransactions(transactions);
       } catch (err) {
         if (currTransaction.handleTransactionError) {
+          this.setState({ step: 0, disableButton: false });
```
am9u: Should this be outside the `if` block? That way the buttons resets even if the transaction doesn't have a `handleTransactionError`.

Though I think all transactions buttons w/ the modal have `handleTransactionError` set, but lower level implementations of `TransactionButton` might not have that set.

nickreynolds: good point, i'll take it out of the if

PR #231, opened 2018-06-22
Bug fixes and improvements for DApp UI

am9u:
- Fix the Request Appeal button to actually work
- Fix formatting of the Challenge Results graph (will be DRY-ing this up in an upcoming update)
- Update the Salt form inputs for Commit/Reveal vote to indicate that Salt should be a number

### DIFF
--- a/packages/components/src/ListingDetailPhaseCard/ChallengeResults.tsx
+++ b/packages/components/src/ListingDetailPhaseCard/ChallengeResults.tsx
@@ -1,7 +1,6 @@
 import * as React from "react";
 import styled, { StyledComponentClass } from "styled-components";
 import { colors } from "../styleConstants";
-import { getNumberStringWithCommaDelimeters } from "@joincivil/utils";
 import { buttonSizes, DarkButton } from "../Button";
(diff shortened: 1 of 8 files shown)


packages/components/src/ListingDetailPhaseCard/ChallengeResults.tsx
```diff
@@ -83,7 +82,7 @@ export class ChallengeResults extends React.Component<ChallengeResultsProps> {
             <VotesPerTokenVote vote="remain">
               <span>✔</span> Remain
             </VotesPerTokenVote>
-            <VotesPerTokenCount>{getNumberStringWithCommaDelimeters(this.props.votesAgainst)} CVL</VotesPerTokenCount>
```
nickreynolds: should this just use `getFormattedTokenBalance`?

am9u: Yeah, I removed this method call... and `getFormattedTokenBalance` gets called in the views that implement this view. I'm going to optimize and move those helper calls to this view, but I think I'll do that in a near-future update.

PR #236, opened 2018-06-25
Fixing Challenge Stage Transitions

nickreynolds:
- Fixing callbacks so listings re-render properly when changing state without event being fired. Adding expiry to some components' props so they re-render when it changes.


packages/dapp/src/helpers/listingEvents.ts
```diff
@@ -9,6 +9,7 @@ import { Observable, Subscription } from "rxjs";
 import { BigNumber } from "bignumber.js";
 
 const listingTimeouts = new Map<string, number>();
+const setTimeoutTimeouts = new Map<string, number>();
```
am9u: This is just a funny name

PR #238, opened 2018-06-27
DApp: Detailed progress modals for transactions

am9u:
Update progress modals for all the transactions with better messaging about which transaction is currently in progress


packages/components/src/ListingDetailPhaseCard/ChallengeRevealVoteCard.tsx
```diff
@@ -42,14 +44,37 @@ export class ChallengeRevealVoteCard extends React.Component<
...
-            onInputChange={this.props.onInputChange}
-            transactions={this.props.transactions}
-          />
-        </StyledListingDetailPhaseCardSection>
+        <StyledListingDetailPhaseCardSection>{this.renderCommitVote()}</StyledListingDetailPhaseCardSection>
       </StyledListingDetailPhaseCardContainer>
     );
   }
+
+  private renderCommitVote = (): JSX.Element => {
```
nickreynolds: something about this function's name feels wrong.  should this be "renderRevealVote"

PR #242, opened 2018-06-29
Update lcov-result-merger dependency to support node v10

tobek:
lcov-result-merger v2 required node engine <=9

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   "devDependencies": {
     "coveralls": "^3.0.0",
     "ganache-cli": "^6.1.0-beta.4",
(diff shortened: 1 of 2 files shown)


ritave:
@tobek I've played with the node 10 and lcov, but seems more dependencies don't work yet which I didn't have time to debug. I'm closing this for now and we should go back when we need it / packages are more stable

tobek:
Ok good to know, makes sense. I ran into this while testing setting up repo from scratch as a first-time visitor/contributor coming in via blog posts - this was the only snag preventing a hands-free install and build. But I think most potential contributors will be savvy enough to handle it!

PR #246, opened 2018-07-02
Newsroom manager should check for validity [ch1304]

walfly:


packages/components/src/input/AddressInput.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { storiesOf } from "@storybook/react";
+import * as React from "react";
+import { AddressInput } from "./AddressInput";
+
+storiesOf("AddressInput", module)
```
am9u: Should we put all the form inputs under one "Input" story?

Not a blocker... just a question.

PR #248, opened 2018-07-02
Implement Listing Details Header design

am9u:
This update moves the Listing Details header design into `components` and includes some refactoring to DRY up some of the for getting listing data.


packages/dapp/src/components/listinglist/ListingList.tsx
```diff
@@ -25,7 +25,7 @@ class ListingList extends React.Component<ListingListOwnProps> {
         {this.props.challenges &&
           this.props.challenges.map(c => {
             index++;
-            return <ListingListItem key={c} challengeID={c!} even={index % 2 === 0} />;
+            return <ChallengeListingListItem key={c} challengeID={c!} even={index % 2 === 0} />;
```
nickreynolds: I think we can get rid of the `even` prop now that it's cards, right?

am9u: The table view for the Registry page -- which @sruddy is working on -- still has zebra styling, which is why I left that prop in.

PR #249, opened 2018-07-03
listing history redux

nickreynolds:
- improve the way we start and store subscriptions of listing history events
- unsubscribe from listing history subscriptions when network changes


packages/dapp/src/actionCreators/network.ts
```diff
@@ -1,12 +1,24 @@
...
   return {
     type: networkActions.SET_NETWORK,
     data: network,
   };
 };
+
+export const setNetwork = (network: string): any => {
+  return (dispatch: any, getState: any): AnyAction => {
+    const { listingHistorySubscriptions } = getState().networkDependent;
+    listingHistorySubscriptions.forEach((subscription: Subscription, key: string) => {
```
walfly: shouldn't the linter complain about the unused argument?

packages/dapp/src/actionCreators/network.ts
```diff
@@ -1,12 +1,24 @@
...
+      subscription.unsubscribe();
```
walfly: should the listing also be removed from the map?

packages/dapp/src/reducers/listings.ts
```diff
@@ -67,6 +68,18 @@ export function histories(
   }
 }
 
+export function listingHistorySubscriptions(
+  state: Map<string, Subscription> = Map<string, Subscription>(),
+  action: AnyAction,
+): Map<string, Subscription> {
+  switch (action.type) {
+    case listingActions.ADD_HISTORY_SUBSCRIPTION:
+      return state.set(action.data.address, action.data.subscription);
```
walfly: theres no check to make sure you don't overwrite an already existing subscription.  That could cause memory leaks.

packages/dapp/src/actionCreators/network.ts
nickreynolds: when the SET_NETWORK action is processed in the network reducer, everything is cleared out, including this map

packages/dapp/src/reducers/listings.ts
nickreynolds: I check in the action creator before dispatching. What's the right thing to do here in the reducer? I thought I can't really do anything like unsubscribe from an existing subscription in the reducer since reducers can't have side effects, right?

walfly: Checking in the action is fine, i missed that.  I would have only set it if it was not already set but they're the same.

packages/dapp/src/actionCreators/network.ts
nickreynolds: it doesn't seem to, and i'm going to leave it as is because it makes it more clear that it's a map instead of a list/set. Although it's absolutely crazy that it's (value, key) and not (key, value)

PR #251, opened 2018-07-03
DApp: Build Submit Challenge modal

am9u:


packages/dapp/src/components/listing/ListingPhaseActions.tsx
```diff
@@ -167,41 +212,41 @@ class ListingPhaseActions extends React.Component<ListingPhaseActionsProps> {
...
   }
 
+  private handleSubmitChallenge = () => {
+    this.setState({ isChallengeModalOpen: true });
+  };
+
+  private updateChallengeStatement = (value: any) => {
+    // Using functions here ensures that these state changes
+    // happen synchronously
+    this.setState(() => ({ isPostStatementDisabled: true }));
```
nickreynolds: what is `isPostStatementDisabled` used for? this function is kinda confusing

am9u: Ah, that state value isn't being used anymore. I'll remove that code.

The 'Post Statement' and 'Submit Challenge' actions were initially separate buttons (per the mockup), and that value toggled the `disabled` attribute on the 'Post Statement' button.

PR #255, opened 2018-07-09
Added a backported websocket provider implementaton for web3 0.x

ritave:
Caveat - The provider never closes and keeps the Node process forever running. This is consistent with the upstream implementation


packages/dapp/src/components/Main.tsx
```diff
@@ -31,7 +31,7 @@ class Main extends React.Component<DispatchProp<any> & RouteComponentProps<any>>
 
   public onNetworkUpdated = async (): Promise<void> => {
     const civil = getCivil();
-    if (civil.network) {
+    if (civil.network === "4") {
       this.props.dispatch!(setNetwork(civil.network));
```
nickreynolds: I think we need to set the network even if it's not rinkeby so that we properly render the error ("switch to rinkeby") bar

packages/components/src/DetailTransactionButton.tsx
```diff
@@ -95,9 +102,41 @@ export class DetailTransactionButton extends React.Component<
...
   public async componentDidMount(): Promise<void> {
+    this.createEthereumSubscription(this.props.civil);
     await this.divinePrice(this.props.estimateFunctions);
   }
 
+  public async componentWillUnmount(): Promise<void> {
+    this.unbscribeEthereum();
+  }
+
+  public createEthereumSubscription(civil?: Civil): void {
```
nickreynolds: this stuff is good but it should all happen outside of the component and be stored in redux, like how we store network/account in redux currently

packages/core/src/templates/contract/partials/constructor.handlebars
```diff
@@ -37,8 +36,9 @@ public static deployTrusted = {
        * gets into mempool, and second when it's mined. But it's called only once, so we have to resolve
        * the contract on our own
        */
+      console.log(options);
       clazz.new(
-        {{> params inputs=contract.inputs}}
+        {{> params inputs=ctor.inputs}}
```
nickreynolds: not sure why this changed

packages/components/src/DetailTransactionButton.tsx
ritave: @nickreynolds There is no redux in the components package and it's used in multiple places, not only the dapp. This is the most backwards-compatible way I've found. If we get eg a standard of how we should write react components and refactor all our front-ends then sure
cc @dankins we really need one

nickreynolds: well you would just pass this data in to the component as a prop and the DApp would get it from its redux

PR #258, opened 2018-07-10
DApp: Registry Tabs and Sub Tabs

sruddy:


packages/dapp/src/components/utility/Tabs.tsx
```diff
@@ -1,17 +1,32 @@
 import * as React from "react";
```
am9u: Should these components be in `components` instead of `dapp`, so the tab functionality is shared and also testable in Storybook?

packages/dapp/src/components/utility/Tabs.tsx
```diff
@@ -1,17 +1,32 @@
...
-  padding: 0;
+const TabContainer = styled.div`
+  display: flex;
+  justify-content: center;
+  margin: 0 auto;
 `;
 
+const TabNav = styled.div`
+  margin: 0 auto 50px;
+  &.listings-nav {
```
am9u: If we need DApp-specific styles, I think we can import the `Tabs` component from `@joincivil/components` into the DApp and extend it.

Something like this should work.

```
const ListingsNavTabs = Tabs.extend`
  <specific styles here>;
`;

<ListingsNavTabs>
  <Tab />
</ListingNavTabs>
```

packages/dapp/src/components/utility/Tabs.tsx
```diff
@@ -1,17 +1,32 @@
...
+    background-color: #e9e9ea;
```
am9u: Is this color not in `@joincivil/components`? 

If it isn't, maybre we should use the closest grey that we have defined.

packages/components/src/Tabs.tsx
```diff
@@ -0,0 +1,117 @@
...
+    });
+  };
+
+  // Render current active tab content
+  public renderActiveTabContent = (): JSX.Element => {
+    const { children } = this.props;
+    const { activeTabIndex } = this.state;
+
+    if (children![activeTabIndex]) {
+      return children![activeTabIndex].props.children;
```
walfly: doesn't typescript freak out at you for this?

sruddy:
@am9u your changes are updates

packages/components/src/Tab.tsx
```diff
@@ -33,12 +34,19 @@ StyledLi.defaultProps = {
...
+`;
+
 export class Tab extends React.Component<TabProps> {
   public render(): JSX.Element {
     const TabComponent = this.props.TabComponent || StyledLi;
+    const hasCount = this.props.tabCount ? <Count>{this.props.tabCount}</Count> : "";
+
     return (
       <TabComponent isActive={this.props.isActive} onClick={this.onClick}>
         {this.props.title}
```
am9u: Instead of supporting a `tabCount` prop, should we just let `title` be either a string or a JSX.Element? 

This way the implementer has control over what gets rendered and we don't have to worry about supporting another property for this component.

packages/components/src/Tabs.stories.tsx
```diff
@@ -3,21 +3,109 @@ import * as React from "react";
...
+  }
+`;
+
+const ListingsSubnavTabs = styled.div`
+  margin: 30px auto 50px;
+  width: 100%;
+`;
+
+const ListingsSubnavTab = styled.li`
+  background-color: ${(props: TabComponentProps) => (props.isActive ? "#e9eeff" : "transparent")};
```
am9u: Is `#e9eeff` in our `colors` constant? If not, can we use something close to it instead?

packages/components/src/Tabs.stories.tsx
```diff
@@ -3,21 +3,109 @@ import * as React from "react";
...
-  font-family: ${props => props.theme.sanserifFont};
-  border-bottom: ${(props: TabComponentProps) => (props.isActive ? "3px solid red" : "none")};
+  line-height: 17px;
+  padding: 20px 44px;
+  &:last-of-type {
+    border-right: 1px solid ${colors.accent.CIVIL_GRAY_4};
+  }
+  &:hover {
+    color: ${colors.accent.CIVIL_BLUE};
+    background-color: #e9eeff;
```
am9u: Same color comment as above

packages/components/src/Tabs.stories.tsx
am9u: Actually, this is just in Storybook, right?

Then this is probably ok.

packages/components/src/Tabs.tsx
```diff
@@ -6,18 +6,24 @@ export interface TabsProps {
   activeIndex?: number;
   children: Array<React.ReactElement<TabProps>>;
   TabComponent?: any;
+  TabsNavComponent?: any;
```
am9u: Should this type be `React.StatelessComponent`?

packages/components/src/Tabs.tsx
```diff
@@ -6,18 +6,24 @@ export interface TabsProps {
...
 }
 
 export interface TabsState {
   activeIndex: number;
 }
 
-const StyledUl = styled.ul`
+const StyledNav = styled.nav`
+  border-bottom: 1px solid #d8d8d8;
```
am9u: Let's use a color from our contstants file here.

packages/components/src/Tabs.tsx
```diff
@@ -6,18 +6,24 @@ export interface TabsProps {
...
+`;
+
+const TabContainer = styled.ul`
+  display: flex;
   list-style: none;
+  margin: 0 auto;
+  max-width: 1200px;
```
am9u: Should we have a `max-width` set at the component level?

I'm thinking we rely on `width: 100%` and have the parent container specify the width. That should make it more reusable.

packages/dapp/src/components/listinglist/styledComponents.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from "react";
+import styled, { StyledComponentClass } from "styled-components";
+import { colors, fonts, TabComponentProps } from "@joincivil/components";
+
+export const ListingsTabNav = styled.div`
```
am9u: One more thought...

Should these styles be in `@joincivil/components`?

They are DApp-specific so I can see why they live here, but our components package is pretty much our dapp and our plugin styles. And if future devs or designers update these styles, they would currently have to update them in both the Storybook stories as well as the DApp.

Maybe restoring the `tabs` directory in `@joincivil/components` is the way to go, and we can put the functional components in there, along with these styles, and just export everything from the directory.

PR #259, opened 2018-07-10
Added Salt Input to Vote Reveal

jorgelo:
* Adds a special input box called SaltInput.

### DIFF
--- a/packages/components/src/input/SaltInput.tsx
+++ b/packages/components/src/input/SaltInput.tsx
@@ -1,0 +1,55 @@
+import * as React from "react";
+import { TextInput } from "../input/";
+import { wordsToSalt } from "@joincivil/utils";
+
+export interface SaltInputProps {
+  name: string;
+  label?: string;
+  placeholder?: string;
+  onChange(name: string, value: string): void;
+}
+
+export interface SaltInputState {
+  isValid: boolean;
+  value: string;
+}
+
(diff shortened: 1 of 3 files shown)


packages/components/src/input/SaltInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public render(): JSX.Element {
+    const { name, label, placeholder } = this.props;
+    const { value, isValid } = this.state;
+
+    return <TextInput label={label} placeholder={placeholder} name={name} onChange={this.handleChange} value={value} />;
+  }
+
+  private handleChange(name: string, value: string): void {
```
am9u: You could probably do this and get rid of the `this.handleChange.bind()` line in the constructor.

(Which would match our coding style thusfar).

```
private const handleChange = (name: string, value: string): void => {
 ...
}
```

am9u:
Looks good!

Just one small comment.

packages/components/src/input/SaltInput.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from "react";
+import { TextInput } from "../input/";
+import { wordsToSalt } from "@joincivil/utils";
+
+export interface SaltInputProps {
+  name: string;
```
nickreynolds: why "name"?

PR #263, opened 2018-07-11
theme buttons

walfly:
I think this pattern is pretty good for letting us reuse and retheme things.  added everyone as a reviewer to share.


packages/components/src/Button.tsx
```diff
@@ -74,70 +92,82 @@ const BaseButton = styled(ButtonComponent)`
...
-    border: 1px solid ${colors.accent.CIVIL_GRAY_1};
-    color: ${colors.basic.WHITE};
+  background-color: ${props => props.theme.darkButtonBackground};
+  color: ${props => props.theme.darkButtonColor};
+  &:hover, &.active {
+    background-color: ${props => props.theme.darkButtonHoverBackground};
   }
 `;
 
-export const CancelButton = SecondaryButton.extend`
```
am9u: This looks cool to me. 

My only comment is that I am using `CancelButton` in the DApp now, because it was defined and available.

I don't necessarily need to use it, but we should have a `TextButton` for cases where we have a button but the style looks like plain text or a link.

walfly: I only deleted it because i searched and no one was using it

packages/components/src/Button.tsx
```diff
@@ -22,6 +22,24 @@ export interface ButtonProps {
   onClick?(ev: any): void;
 }
 
+export interface ButtonTheme {
+  primaryButtonBackground?: string;
+  primaryButtonColor?: string;
+  primaryButtonHoverBrackground?: string;
```
tobek: typo!

PR #267, opened 2018-07-12
Add Challenge Statement summary field to Submit Challenge modal

am9u:


packages/components/src/SubmitChallengeModal.tsx
```diff
@@ -152,8 +162,20 @@ export class SubmitChallengeModal extends React.Component<
     return (
       <StepStyledFluid index={this.props.index || 0}>
         <ModalSectionHeader>State reasons for your challenge</ModalSectionHeader>
+        <CopyLarge>Enter a summary of the reasons for your challenge (Max 120 characters)</CopyLarge>
+        <SectionFormOuter>
+          <FormInputGroup>
+            <RichTextEditor
```
nickreynolds: the summary should be plaintext, not richtext, imo

packages/components/src/SubmitChallengeModal.tsx
```diff
@@ -196,4 +257,28 @@ export class SubmitChallengeModal extends React.Component<
       </StepStyledFluid>
     );
   };
+
+  private handleBeforeInput = () => {
```
nickreynolds: I think these functions should go on a TextInput component

am9u:
Sure, that's actually much simpler.

PR #269, opened 2018-07-12
Components Tooltip [ch1314]

walfly:


packages/components/src/QuestionToolTip.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  constructor(props: QuestionToolTipProps) {
+    super(props);
+    this.divEl = null;
+    this.state = {
+      open: false,
+    };
+  }
+
+  public componentDidMount(): void {
```
am9u: Do we need to worry about removing `bucket` when the component unmounts?

walfly: yeah, i guess thats a good thing to do

PR #270, opened 2018-07-13
Add `try/catch` to Challenge Statement get method

am9u:


packages/core/src/contracts/tcr/challenge.ts
```diff
@@ -66,7 +66,12 @@ export class Challenge {
   private async getChallengeStatement(): Promise<ContentData | undefined> {
     const uri = await this.getChallengeURI();
     if (uri) {
-      return this.contentProvider.get({ uri, contentHash: "" });
+      try {
+        const challengeStatement = await this.contentProvider.get({ uri, contentHash: "" });
+        return challengeStatement;
+      } catch (e) {
+        return;
```
nickreynolds: probably worth logging something..

PR #273, opened 2018-07-16
Respond to Audit Finding 3.21

nickreynolds:
- Using standard Paramterizer implementaiton and extending it instead of copying + modifying
- will point base to master once PR for finding 3.22 is merged in (since this includes that)


nickreynolds:
yeah, didn't seem worth the modifications to keep it

PR #277, opened 2018-07-18
Load salt from local storage on the reveal

jorgelo:
* On reveal, load salt from localstorage.


nickreynolds:
jorge's offline but this seems good to go and I wanna see it on the dapp

PR #291, opened 2018-07-24
Fix CircleCI cache not updating after subpackages change

### LINKED ISSUE #290: CircleCI cache doesn't take subpackages into account
When a package.json inside `packages/*` change, the cache key doesn't change, meaning CircleCI builds use outdates `node_modules/` folder

ritave:
fix #290


ritave:
Will be fixed by #289

PR #293, opened 2018-07-24
Walfly/add tabs

walfly:


am9u:
Is there any overlap here with @sruddy's Tab PR?

https://github.com/joincivil/Civil/pull/258

sruddy:
A lot of overlap but most of my changes are styles. If you need to get this out go for it. I should be back to working on the dapp in the next day and will get mine integrated asap. My PR already has the ability to pass in different components like @am9u mentioned.

walfly:
Should have checked if anyone was working on tabs, yeah we need them right away. will add an option to pass in a tab type

walfly:
would we ever want individual components per tab?

am9u:
what do you mean by individual components per tab @walfly?

walfly:
@am9u I meant like would you want to pass a different tabcomponent into each tab, or always the same for all tabs.  seems like it should always be the same

PR #309, opened 2018-07-27
require owner for revising the charter

walfly:


packages/contracts/test/newsroom/newsroom.ts
```diff
@@ -221,6 +221,18 @@ contract("Newsroom", (accounts: string[]) => {
         contentId = event!.args.contentId;
       });
 
+      it("can't be the charter if you're not an owner", async () => {
```
nickreynolds: "can't update" not "can't be" right?

walfly: lol yeah

PR #319, opened 2018-07-30
abstract out tool tip, add date to signatures

walfly:


packages/components/src/ToolTip.tsx
```diff
@@ -48,6 +49,7 @@ const Tip = styled.div`
     border-top: 10px solid rgba(21, 21, 21, 0.9);
   }
 `;
+// z-index to compete with wp
```
am9u: sigh.

packages/components/src/QuestionToolTip.tsx
```diff
@@ -1,115 +1,12 @@
...
-
-  public componentWillUnmount(): void {
-    this.bucket.remove();
-  }
-
+export class QuestionToolTip extends React.Component<ToolTipProps> {
   public render(): JSX.Element {
-    let tip = null;
-    let hitBox = null;
     const color = this.props.disabled ? colors.accent.CIVIL_GRAY_3 : "#000";
```
am9u: Could probably use `colors.primary.BLACK` instead of `#000`

PR #322, opened 2018-07-31
Various updates for DApp + Components

am9u:
- Remove lorem ipsum description and owner from Listing Summary cards on
the Registry view (which fixes the wrapping issue for newsroom name)
- Add and render additional Listing Phase Labels
- Add 'Back to Registry' link to Listing Detail page
- Add basic routing, and updated browser history, when navigating
between listing tabs
- Minor nav bar styles tweaks
- Fix bug that prevented Submit Challenge modal from displaying


am9u:
@sruddy Oops! Sorry, I was just adding links back to the Registry page from the Details page and went down the rabbit hole. I saw that you're working on Rejected cards so tried to stay away from those.

PR #324, opened 2018-07-31
DApp: Add Challenge Results to Rejected listing cards

sruddy:


packages/components/src/ListingSummary/ListingSummaryRejected.tsx
```diff
@@ -0,0 +1,52 @@
...
+  StyledListingSummarySection,
+  StyledListingSummaryNewsroomName,
+  ChallengeResultsContain,
+  NewsroomIcon,
+} from "./styledComponents";
+
+export interface ListingSummaryRejectedComponentProps {
+  address?: EthAddress;
+  name?: string;
+  owners?: EthAddress[];
```
am9u: I think we can get rid of `owners`

am9u: And probably `description` as well, since the Challenge Results replace the description, right?

PR #331, opened 2018-08-03
Respond to Audit Finding 3.1 and 3.4

ritave:
Added replay attack safety for Newsroom smart-contract

### DIFF
--- a/packages/contracts/contracts/newsroom/Newsroom.sol
+++ b/packages/contracts/contracts/newsroom/Newsroom.sol
@@ -33,6 +33,13 @@ contract Newsroom is ACL {
   string private constant ROLE_EDITOR = "editor";
 
   mapping(uint => Content) private contents;
+  /*
+  Maps the revision hash to original contentId where it was first used.
+  This is used to prevent replay attacks in which a bad actor reuses an already used signature to sign a new revision of new content.
+  New revisions with the same contentID can reuse signatures by design -> this is to allow the Editors to change the canonical URL (eg, website change).
+  The end-client of those smart-contracts MUST (RFC-Like) verify the content to it's hash and the the hash to the signature.
+  */
+  mapping(bytes32 => UsedSignature) private usedSignatures;
 
   /**
   @notice The number of different contents in this Newsroom, indexed in <0, contentCount) (exclusive) range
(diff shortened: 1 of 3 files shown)


thec00n:
Just having a look at this. Quick question have you considered using EIP712?

ritave:
@thec00n Yeah, I'm very excited about that EIP, but until `eth_signTypedData` is supported by at least geth, metamask and ganache I'd rather play it safe.

packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -184,27 +191,44 @@ contract Newsroom is ACL {
     require(content.author == 0x0 || content.author == author);
     require(content.revisions.length > revisionId);
 
+    if (contentId == 0) {
```
thec00n: Is there anything special about contents[0]?

packages/contracts/contracts/newsroom/Newsroom.sol
```diff
@@ -184,27 +191,44 @@ contract Newsroom is ACL {
...
 
-  function verifyRevisionSignature(address author, Revision storage revision) view internal returns (bool isSigned) {
+  function verifyRevisionSignature(address author, uint contentId, Revision storage revision) internal returns (bool isSigned) {
     if (author == 0x0 || revision.signature.length == 0) {
       require(revision.signature.length == 0);
       return false;
     } else {
+      // The url is is not used in the cryptography by design,
+      // the rationale is that the Author can approve the content and the Editor might need to set the url
+      // after the fact, or things like DNS change, meaning there would be a question of canonical url for that article
```
thec00n: I was under the impression that the contentUri is an IPFS hash. The comments indicate that this might change in the future allowing content from URLs?

ritave: @thec00n The Uri is whatever that specific Newsroom wants it to be. We're making multiple URI solutions, with basics such as HTTP, IPFS and more experimental such as using Event Log of a Smart Contract or appended data at the end of the transaction

packages/contracts/contracts/newsroom/Newsroom.sol
ritave: @thec00n Yes, it's a newsroom charter, which is established during construction of a newsroom. That article is Owner-editable only and constitutes a "About page".

I shouldn't have added this change to this PR though. My bad

packages/contracts/contracts/newsroom/Newsroom.sol
thec00n: For unsigned and signed content any Civil client needs to download the content at contentUri and generate a hash and verify that the hash matches the content that is stored at contentUri. It needs to be ensured by the Civil client that no invalid content is displayed to the user because by design a user could add a contentUri where the contentHash does not match its content. This is could be a nuisance and used to attack the Civil system.

ritave: @thec00n I agree, this is a trade-off we have to do due to content being outside of Ethereum.

thec00n: Ok, just want to make sure that you guys are aware that this could be a problem.

ritave:
@thec00n Did anyone grab your attention or would say this good to go to be merged?

thec00n:
I think you should address point 1. Points 2 and 3 are comments to make sure you are aware:

* The information in the UsedSignature struct is unnecessary to prevent signature replay attacks. Just storing the contentHash in a dynamic array or mapping is sufficient https://github.com/joincivil/Civil/blob/1e598ebf4eb28979cfa8faee44997df11054eb4f/packages/contracts/contracts/newsroom/Newsroom.sol#L274-L277 Also currently the contentHash is not stored in the usedSignatures mapping. The contentHash needs to be stored after successful verification in signRevision() and pushRevision().
* It needs to be ensured by the Civil client that no invalid content is displayed to the user because by design a user could add a contentUri where the contentHash does not match its content. This is could be a nuisance and used to attack the Civil system.
* Front running is an issue in the current design. So if author A creates a piece of content and sends a revision update, author B could listen on the network and sends a transaction with the same contentHash and higher gas amount. User B could claim that the content is from him. That's not a high severity issue in the case of Civil I would think but still something to consider.

PR #338, opened 2018-08-03
Respond to Audit Finding 3.13

nickreynolds:
- Copying latest standard implementation of Registry into AddressRegistry, updating it to use addresses instead of bytes32, updating visibility of some functions so they can be overwritten
- Fixing tests after changes

### DIFF
--- a/packages/contracts/contracts/tcr/CivilTCR.sol
+++ b/packages/contracts/contracts/tcr/CivilTCR.sol
@@ -65,16 +65,15 @@ contract CivilTCR is RestrictedAddressRegistry {
   @param paramsAddr Address of a Parameterizer contract
   @param govt IGovernment contract
   */
-  function initTCR(
+  constructor(
     address tokenAddr,
     address plcrAddr,
     address paramsAddr,
     IGovernment govt
-  ) public
+  ) public RestrictedAddressRegistry(tokenAddr, plcrAddr, paramsAddr, "CivilTCR")
   {
-    init(tokenAddr, plcrAddr, paramsAddr, "CivilTCR");
     civilVoting = CivilPLCRVoting(plcrAddr);
-    require(address(govt) != 0 && address(government) == 0);
+    require(address(govt) != 0);
     require(govt.getGovernmentController() != 0);
     government = govt;
   }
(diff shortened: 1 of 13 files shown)


packages/contracts/contracts/tcr/CivilTCR.sol
```diff
@@ -56,21 +58,23 @@ contract CivilTCR is RestrictedAddressRegistry {
...
     address tokenAddr,
     address plcrAddr,
     address paramsAddr,
     IGovernment govt
-  ) public RestrictedAddressRegistry(tokenAddr, plcrAddr, paramsAddr)
+  ) public
   {
-    require(address(govt) != 0);
+    init(tokenAddr, plcrAddr, paramsAddr, "CivilTCR");
+    civilVoting = CivilPLCRVoting(plcrAddr);
```
ritave: Nitpick: duplicates the same address, might desync with further changes in the code if not careful.
Using `CivilPLCRVoting(voting)` everywhere should work too?

packages/contracts/contracts/tcr/CivilTCR.sol
```diff
@@ -56,21 +58,23 @@ contract CivilTCR is RestrictedAddressRegistry {
...
-  @notice Contructor Sets the addresses for token, voting, parameterizer, appellate, and fee recipient
+  @notice Init function calls AddressRegistry init then sets IGovernment
   @dev passes tokenAddr, plcrAddr, paramsAddr up to RestrictedAddressRegistry constructor
   @param tokenAddr Address of the TCR's intrinsic ERC20 token
   @param plcrAddr Address of a PLCR voting contract for the provided token
   @param paramsAddr Address of a Parameterizer contract
   @param govt IGovernment contract
   */
-  function CivilTCR(
+  function initTCR(
```
ritave: Why is this not a constructor anymore? All other functions in this smart-contract would have to verify if it it was initialised beforehand to be able to work with known state.

Please make it into a constructor again

nickreynolds: this brings it in line with the standard implementation. asking consensys TCR to explain reasoning behind init vs constructor, will get back to you

packages/contracts/contracts/tcr/CivilTCR.sol
nickreynolds: that would require modifying the AddressRegistry code more, which is what we're trying to avoid

PR #347, opened 2018-08-08
styled anchor tag if you pass an href to button

walfly:


walfly:
close cause toby also did it

PR #349, opened 2018-08-08
add metamask modal

walfly:


packages/components/src/types.d.ts
```diff
@@ -1,3 +1,4 @@
 declare module "@storybook/addon-storyshots";
 declare module "storybook-react-router";
 declare module "react-rte";
+declare module "*.png";
```
walfly: @sruddy fixes the file loader

PR #354, opened 2018-08-09
UI updates for Voting phase Listing Detail cards

am9u:


packages/components/src/ToolTip.tsx
```diff
@@ -52,6 +52,7 @@ const Tip = styled.div`
 // z-index to compete with wp
 
 const Wrapper = styled.div`
+  display: inline-block;
```
am9u: @walfly This update won't mess with your stuff, will it?

PR #355, opened 2018-08-09
Dapp: Registry hero component

sruddy:


packages/components/src/Hero/HomepageHero.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export class HomepageHero extends React.Component {
+  public render(): JSX.Element {
+    return (
+      <>
+        <HeroLabel>Civil Registry</HeroLabel>
+        <HeroHeading>
+          The Civil Registry is a whitelist of community-approved newsrooms that have publishing rights on Civil.
+        </HeroHeading>
+        <HeroBlockTextLink href="/home">Learn how to participate in our governance</HeroBlockTextLink>
```
am9u: Can we make the `/home` url a prop that gets passed into the component?

packages/components/src/Hero/HomepageHero.tsx
```diff
@@ -0,0 +1,21 @@
...
+        <Button size={buttonSizes.MEDIUM} to="/home">
+          APPLY TO JOIN REGISTRY
+        </Button>
+        <HeroSmallText>1,000 CVL required to apply</HeroSmallText>
```
am9u: `1,000 CVL` should probably be a prop too... which should probably get hydrated by `minDeposit` from the `parameters` reducer in the DApp.

packages/components/src/Hero/styledComponents.tsx
```diff
@@ -0,0 +1,47 @@
...
+  display: block;
+  font-family: ${fonts.SANS_SERIF};
+  font-size: 20px;
+  font-weight: 800;
+  letter-spacing: -0.27px;
+  margin-bottom: 30px;
+`;
+
+export const HeroHeading = styled.h2`
+  color: ${colors.basic.WHITE};
```
am9u: Should we move the `color` and `font-family` definitions up to a parent element and save a few bytes? Maybe `HeroOuter`?

The build for the dapp is getting big and saving a couple bytes here and there might be nice

packages/components/src/Hero/HomepageHero.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Button, buttonSizes } from "../Button";
+import { HeroLabel, HeroHeading, HeroBlockTextLink, HeroSmallText } from "./styledComponents";
+
+export class HomepageHero extends React.Component {
+  public render(): JSX.Element {
+    return (
+      <>
+        <HeroLabel>Civil Registry</HeroLabel>
+        <HeroHeading>
+          The Civil Registry is a whitelist of community-approved newsrooms that have publishing rights on Civil.
```
nickreynolds: can we move this text to a constants file (which we need to create and start using for all text)

sruddy:
@am9u I made your updates

packages/components/src/Hero/constants.ts
```diff
@@ -0,0 +1,7 @@
+export enum homepageHeroText {
```
am9u: FYI, I think we're going to go with a different way of rendering these text strings. I'm working on it in a PR now, but we can update these strings in the future.

PR #357, opened 2018-08-10
Proof of use controller

ritave:


packages/contracts/contracts/proof-of-use/GroupTokenController.sol
```diff
@@ -0,0 +1,62 @@
...
+  // After a group proves token use, they can add themselves to GLOBAL_GROUP,
+  // allowing to exchange tokens with everyone who already proven too
+  address constant GLOBAL_GROUP = 0x2;
+
+  // The maximal amount of addresses inside one group before being added to global
+  uint public maxGroupSize = 4;
+
+  UnionFind.Data private groups;
+
+  function isUseProven(address a) public returns (bool) {
```
nickreynolds: pretty sure this would need to take in a number of tokens as well, since use is proven of certain amount of tokens (e.g. can transfer 1000 after proving use of 500)

nickreynolds:
this contract just generally makes me a bit nervous since it's fairly complicated, I don't really feel 100% qualified to review it

PR #361, opened 2018-08-11
Responding to Audit Issue 56

nickreynolds:
- Adding voting process for changing government parameters


nickreynolds:
what do we need to do to get more resources to evaluate this change? our deploy date is now Oct 7

PR #366, opened 2018-08-16
core updates

nickreynolds:
- adding event data to UserChallengeData so we have salt and numtokens ready to use in DApp
- removing salt input from claim rewards since we already have it

### DIFF
--- a/packages/core/src/contracts/tcr/voting.ts
+++ b/packages/core/src/contracts/tcr/voting.ts
@@ -5,8 +5,9 @@ import * as Debug from "debug";
 import { Observable } from "rxjs";
 import { Bytes32, EthAddress, PollData, TwoStepEthTransaction } from "../../types";
 import { BaseWrapper } from "../basewrapper";
-import { CivilPLCRVotingContract } from "../generated/wrappers/civil_p_l_c_r_voting";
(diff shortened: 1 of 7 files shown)


packages/core/src/contracts/tcr/voting.ts
```diff
@@ -175,6 +176,22 @@ export class Voting extends BaseWrapper<CivilPLCRVotingContract> {
...
 
+  public async getRevealedVoteEvent(
+    pollID: BigNumber,
+    voter: EthAddress,
+  ): Promise<
+    DecodedLogEntryEvent<CivilPLCRVoting.Args._VoteRevealed, CivilPLCRVoting.Events._VoteRevealed> | undefined
+  > {
+    if (await this.didRevealVote(voter, pollID)) {
+      const reveal = await this.instance
+        ._VoteRevealedStream({ pollID }, { fromBlock: 0 })
```
ritave: No filtering by the voter, it's gonna return any kind of reveal instead of first reveal by that specific voter
``` { pollID, voter }, ...```

nickreynolds: dang good catch

PR #392, opened 2018-08-21
Dapp: User nav drawer

sruddy:


sruddy:
I won't be working on governance for the next week. I have a couple placeholders in here that I will add to a new ticket.

packages/components/src/styleConstants.ts
```diff
@@ -29,4 +29,5 @@ export const fonts = {
   SERIF: `"Spectral", serif`,
   SANS_SERIF: `"Libre Franklin", sans-serif`,
   MONOSPACE: `font-family: “SF Mono”, “Segoe UI Mono”, “Roboto Mono”, “Ubuntu Mono”, Menlo, Courier, monospace;`,
+  COURIER: `Courier, sans-serif`,
```
am9u: You should check with Nguyet, but I think the Monospace font is what she wants to use for addresses and numbers.

At least that's the font-stack she gave for those.

packages/components/src/NavBar/NavBar.tsx
```diff
@@ -101,54 +130,90 @@ const UserAvatar = styled.figure`
...
+            userChallengesVotedOnCount={this.props.userChallengesVotedOnCount}
+            userChallengesStartedCount={this.props.userChallengesStartedCount}
+            buyCvlUrl={this.props.buyCvlUrl}
+          />
+        ) : null}
+      </NavOuter>
+    );
+  }
+
+  private toggle(): void {
```
am9u: If you change this method to use an arrow function, you won't need to bind the scope in the constructor and you can just pass the method as the `onClick` prop, instead of using a arrow method there.

Also, the arrow function seems to be the preferred best practice for React these days.

```
private toggle: void = () => {}
```

packages/dapp/src/components/GlobalNav.tsx
```diff
@@ -8,31 +8,43 @@ export interface NavBarProps {
...
 const GlobalNavComponent: React.SFC<NavBarProps> = props => {
   const shouldRenderErrorBar = props.network !== "4";
   return (
     <>
-      <NavBar balance={props.balance} votingBalance={props.votingBalance} />
+      <NavBar
+        balance={props.balance}
+        votingBalance={props.votingBalance}
+        userAccount={props.userAccount}
+        buyCvlUrl="https://civil.co/cvl/"
```
am9u: Not passing the counts for user challenges/votes?

sruddy: This is part of what I'm adding to the new ticket there's a placeholder in there now. https://app.clubhouse.io/civil/story/2100/dapp-nav-and-drawer-updates

PR #399, opened 2018-08-23
Add archive in transaction to newsroom class

walfly:


packages/utils/src/contracts.ts
```diff
@@ -25,3 +25,18 @@ export function is0x0Address(address: string): boolean {
 export function is0x0Hash(hash: string): boolean {
   return hash === "0x0" || hash === "0x0000000000000000000000000000000000000000000000000000000000000000";
 }
+
+export function estimateRawHex(hex: string): number {
```
nickreynolds: can we add some unit/integration tests to this that compare estimates to actual gas used in transactions?

PR #426, opened 2018-09-11
Implement updated UI and designs for Parameterizer

am9u:


packages/dapp/src/helpers/parameterizer.ts
```diff
@@ -38,20 +49,72 @@ export async function initializeProposalsSubscriptions(dispatch: Dispatch<any>):
...
+    dispatch(addOrUpdateProposal(paramProposal));
   });
 }
+
+async function getNextTimerExpiry(paramProposal: any, dispatch: Dispatch<any>): Promise<number> {
+  let nextExpiry;
+
+  switch (paramProposal.state) {
+    case ParamProposalState.APPLYING:
+      nextExpiry = paramProposal.applicationExpiry.valueOf() / 1000;
```
nickreynolds: this is definitely correct? I would've assume it would be `* 1000` since the contract is in seconds but timeouts are in milliseconds

nickreynolds: or just not divided by anything

am9u: Yeah, these are correct actually. The dapp is using the Parameterizer getters for these expiry values, which each return Dates.

https://github.com/joincivil/Civil/blob/master/packages/core/src/contracts/tcr/parameterizer.ts#L383

This could be cleaned up a bit though... let me make a few changes.

am9u:
@sruddy I think that styled component declaration was just missing styles. Just added a fix.

PR #454, opened 2018-09-19
Block newsroom manager if user is not on contract [ch1895]

tobek:
Disables entire component with error message if you're not on the contract at all.

If you're an editor and not an officer, then everything is disabled except the ability to add/remove other editors.


packages/newsroom-manager/src/Newsroom.tsx
```diff
@@ -166,10 +187,17 @@ class NewsroomComponent extends React.Component<NewsroomProps & DispatchProp<any
   };
 
   public isStepDisabled = (index: number) => {
```
tobek: @walfly I updated this function to disable steps based on current user's role, but then saw that this function isn't actually in use. Before I go digging, should this be removed, or is there another way we're handling this, or is this functionality in flux as you work on the manager steps, or should I go ahead and hook it into steps as they are now?

PR #472, opened 2018-09-25
Added proof of use to migrations

ritave:


packages/contracts/test/proof-of-use/usergroups.ts
```diff
@@ -70,13 +66,17 @@ contract("UserGroups", accounts => {
       await expect(userGroups.setMaxGroupSize(5, { from: accounts[1] })).to.eventually.be.rejectedWith(REVERTED);
     });
 
+    xit("signature from non-owners doesn't work", () => {});
```
nickreynolds: can we fill out these unit tests before merging?

packages/contracts/test/proof-of-use/usergroups.ts
```diff
@@ -70,13 +66,17 @@ contract("UserGroups", accounts => {
...
+
     xit("max group size takes effect on any new inGroup unions", async () => {
       await userGroups.setMaxGroupSize(1);
       await expect(userGroups.allowInGroupTransfers(accounts[0], accounts[1])).to.be.rejectedWith(REVERTED);
       await userGroups.setMaxGroupSize(2);
       await userGroups.allowInGroupTransfers(accounts[0], accounts[1]);
       await expect(userGroups.allowInGroupTransfers(accounts[0], accounts[2])).to.be.rejectedWith(REVERTED);
     });
+
+    xit("can't re-use signatures", () => {});
```
nickreynolds: same here

packages/contracts/contracts/proof-of-use/groups/UserGroups.sol
```diff
@@ -58,28 +70,33 @@ contract UserGroups is OffChainOwnable {
...
     );
 
-    UnionFind.union(groups, globalGroup.parent, group.parent);
+    UnionFind.unionStruct(globalGroup, group);
   }
 
-  function isUseProven(address root) internal view returns (bool) {
-    // TODO(ritave): Connect with TCR
-    return true;
+  function isUseProven(uint usedTokens, uint totalTokens) internal view returns (bool) {
```
nickreynolds: we can't merge this into master until the whitelisting is ready, right? because it would prevent people from applying to TCR via multisig-newsroom

packages/contracts/migrations/13_create_super_group.ts
```diff
@@ -0,0 +1,34 @@
...
+import { POU_SUPER_GROUP } from "../test/utils/constants";
+
+module.exports = (deployer: any, network: string, accounts: string[]) => {
+  const [OPERATOR] = accounts;
+  const signAsync = promisify<string>(web3.eth.sign, web3.eth);
+
+  const CivilTCR = artifacts.require("CivilTCR");
+  const Parameterizer = artifacts.require("CivilParameterizer");
+  const PLCRVoting = artifacts.require("CivilPLCRVoting");
+  const Government = artifacts.require("Government");
```
nickreynolds: I think we could skip adding the government to the supergroup, it doesn't accept tokens

packages/contracts/contracts/proof-of-use/groups/UserGroups.sol
ritave: Whitelisting is only checked during the Token Transfer, since we specifically don't enforce any rules in our ERC20 Token, whitelisting is not in effect. I'd say we should test with whitelisting after we finish all the functionality for users.

packages/contracts/migrations/13_create_super_group.ts
ritave: Great, will do

ritave:
@nickreynolds Whitelisting doesn't prevent anyone from applying to TCRs, this PR will not change anything in the user experience

packages/contracts/migrations/6_dummy_token_sale.ts
```diff
@@ -0,0 +1,12 @@
+import { config } from "./utils";
+
+module.exports = (deployer: any, network: string, accounts: string[]) => {
+  const DummyTokenSale = artifacts.require("DummyTokenSale");
+
+  deployer.then(async () => {
+    if (!config.nets[network] || !config.nets[network].TokenSaleAddress) {
+      console.log("Deploying Dummy Token Sale");
```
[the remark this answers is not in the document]
ritave: Thanks, I'll remove it

PR #478, opened 2018-09-27
upgrade ts config

dankins:


packages/dapp/src/components/listing/ListingHistory.tsx
```diff
@@ -1,12 +1,14 @@
-import { List } from "immutable";
 import * as React from "react";
+import { List } from "immutable";
+
 import { connect, DispatchProp } from "react-redux";
 import { State } from "../../reducers";
 import { getListingHistory } from "../../selectors";
-import ListingEvent from "./ListingEvent";
+// import ListingEvent from "./ListingEvent";
```
ritave: Just remove

packages/dapp/src/components/listing/ListingHistory.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { ListingTabHeading } from "./styledComponents";
-import gql from "graphql-tag";
+
 import { Query } from "react-apollo";
+import gql from "graphql-tag";
```
ritave: I'd love to see any centralized parts of dapp be optional, so that anyone can deploy dapp without using our GraphQL server for speedup. In other words - treat it as a cache layer instead of source of truth

nickreynolds:
this PR is getting replaced by the PRs I'm putting up. The graphql stuff will all be optional (but default to being used)

PR #487, opened 2018-10-03
Migrated Proof of Use

ritave:


packages/contracts/contracts/proof-of-use/groups/UserGroups.sol
```diff
@@ -30,6 +30,13 @@ contract UserGroups is OffChainOwnable, TokenTelemetryI, UnionFind {
   {
   }
 
+  function useProven(address who) external view returns (uint usedTokens, uint totalTokens) {
```
nickreynolds: I think this function could have a slightly clearer name. `usedAndTotalTokensForUser` or something

PR #491, opened 2018-10-03
Import just debounce from lodash (smaller build)

tobek:
@dankins tagging you cause your commit 3039b8206a25dc95cab5b6cc7588d81ea455dc89 set `"allowSyntheticDefaultImports": false` in  components tsconfig. It was wasn't working with type checking for `import debounce from "lodash/debounce"` and making it true didn't break anything else, reading the docs seems like it should be ok.

### DIFF
--- a/packages/components/src/DetailTransactionButton.tsx
+++ b/packages/components/src/DetailTransactionButton.tsx
@@ -1,5 +1,5 @@
 import { Civil, EthAddress } from "@joincivil/core";
-import { debounce } from "lodash";
+import debounce from "lodash/debounce";
 import * as React from "react";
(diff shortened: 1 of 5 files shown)


walfly:
@tobek it breaks the production build because of apollo

walfly:
we'll have to dangle our prayers on tree shaking

tobek:
Ahh ok. `lodash-es` would work, but also I assume that some of our dependencies are already using lodash and if the versions are compatible that might save space too, also there were gnarly issues with multiple lodash's in WP plugin so... anyways will just leave it as it is.

PR #502, opened 2018-10-08
Added Proof of Use to core

ritave:


nickreynolds:
can't merge with artifacts changes during playtest

PR #512, opened 2018-10-09
Added automatic Proof-Of-Use addition to the Newsroom creation flow

ritave:


packages/contracts/contracts/proof-of-use/groups/UserGroups.sol
```diff
@@ -61,7 +61,15 @@ contract UserGroups is OffChainOwnable, TokenTelemetryI, UnionFind {
     UnionFind.unionStruct(UnionFind.findStruct(a), UnionFind.findStruct(b));
   }
 
-  function allowInGroupTransfers(address a, address b) external {
+  // This function is incredibly heavy in terms of gas cost
+  function allowInGroupTransfersAll(address[] members) external {
```
walfly: what would this function be used for?  Just passing an array of addresses instead of a single one? Why have it if its sort of a foot-gun in terms of gas costs?

packages/contracts/contracts/ux/CreateNewsroomInGroup.sol
```diff
@@ -0,0 +1,27 @@
...
+contract CreateNewsroomInGroup {
+  NewsroomFactory factory;
+  UserGroups groups;
+
+  constructor(NewsroomFactory _factory, UserGroups _groups) {
+    factory = _factory;
+    groups = _groups;
+  }
+
+  function create(string name, string charterUri, bytes32 charterHash, address[] initialOwners, uint initialRequired)
```
walfly: This is so that you can create the newsroom and multisig in network in one transaction?  How much more does it cost then doing it in two transactions?

packages/core/src/contracts/newsroom.ts
```diff
@@ -92,7 +93,7 @@ export class Newsroom extends BaseWrapper<NewsroomContract> {
     const account = await requireAccount(ethApi).toPromise();
     const txData: TxData = { from: account };
 
-    const factory = await NewsroomFactoryContract.singletonTrusted(ethApi);
+    const factory = await CreateNewsroomInGroupContract.singletonTrusted(ethApi);
```
walfly: Should this be so baked into the deploy flow if its temporary?

packages/contracts/contracts/ux/CreateNewsroomInGroup.sol
```diff
@@ -0,0 +1,27 @@
...
+    public
+    returns (Newsroom newsroom)
+  {
+    require(initialOwners.length > 0);
+
+    newsroom = factory.create(name, charterUri, charterHash, initialOwners, initialRequired);
+
+    groups.allowInGroupTransfersAll(initialOwners);
```
walfly: I don't intend to ever pass in an array of initial owners.  I don't see a reason to do that.  Its just the generated contracts that need to be added.

PR #518, opened 2018-10-11
Signature server for proof-of-use

ritave:


packages/signing-server/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+    demandOption: true,
+  })
+  .option("nodeEndpoint", {
+    describe: "Ethereum node url",
+    requiresArg: true,
+    demandOption: true,
+  })
+  .strict().argv;
+
+const port = Number.parseInt(args.port, 10);
```
dankins: this should probably be an environment variable, I think NODE_PORT is the most common

packages/signing-server/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+    await civil.accountStream.first().toPromise();
+    const userGroups = await civil.userGroupsSingletonTrusted();
+    const nonce = await userGroups.getMaxGroupSizeUpdateNonce();
+    const signature = actionSigner.signMaxGroupSize(toBuffer(userGroups.address), new BN(nonce), new BN(size));
+    const twoStep: TwoStepEthTransaction = await userGroups.setMaxGroupSize(Number.parseInt(size, 10), signature);
+    ctx.body = twoStep.txHash;
+  }),
+);
+
+app.listen(port);
```
dankins: this needs to bind to 0.0.0.0, and should be configurable by the NODE_HOST env var

packages/signing-server/src/server.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!isValidAddress(groupA) || !isValidAddress(groupB)) {
+      throw new Error("Invalid address");
+    }
+    // Due to interactions with core, web3 doesn't have a default account when trying to send tx
+    // Ensure it's set for the first time
+    await civil.accountStream.first().toPromise();
+    const userGroups = await civil.userGroupsSingletonTrusted();
+    const signature = actionSigner.signUnion(toBuffer(userGroups.address), toBuffer(groupA), toBuffer(groupB));
+    const twoStep: TwoStepEthTransaction = await userGroups.forceUnion(groupA, groupB, signature);
+    ctx.body = twoStep.txHash;
```
dankins: can we wrap this in an object so it is returned as JSON instead of just a string?

ritave: what structure would it be? `{ hash: "hash here" }`?

packages/signing-server/src/server.ts
ritave: Yargs enables env parsing.
`SIGNING_SERVER_PORT=1234` will work, similar for all other params

packages/signing-server/src/server.ts
dankins: yea, maybe { "txHash": "hash here"} to make it more explicit as to what the hash is

PR #522, opened 2018-10-12
 Pre-populate logo and newsroom URL from props if available

tobek:
Also replace un-implemented CMS media library for logo with just an img tag


tobek:
@walfly the designs called for a "upload logo to WP Media Library" button that it's not worth implementing before launch. Was just going to leave the markup in for when we do later. There's also a bunch of styles for it.

But yeah I'll remove it all, I can always blame my way to this commit if I want to get it back. Actually will link this PR to the ticket for implementing it.

PR #539, opened 2018-10-17
Validate form data in charter and roster

tobek:


packages/utils/src/civilHelpers.ts
```diff
@@ -47,6 +47,25 @@ export function getDefaultFromBlock(network: number = 4): number {
   return defaultFromBlocks[network.toString()];
 }
 
+export function isValidHttpUrl(url: string): boolean {
+  if (url.indexOf("http") !== 0) {
+    return false;
+  }
+
+  if (!window.URL) {
```
walfly: theres like a million regexes we could copy from stack overflow for this

PR #545, opened 2018-10-19
Go through every help link in plugin and link to appropriate question on zendesk

NickHimmel:


packages/newsroom-manager/src/Newsroom.tsx
```diff
@@ -65,6 +65,7 @@ export interface NewsroomProps {
   showWalletOnboarding?: boolean;
   showWelcome?: boolean;
   helpUrl?: string;
+  baseUrl?: string;
```
tobek: Should be `helpUrlBase` or something, same with the props on `<Welcome>` and `<WalletOnboarding>`

tobek:
I'd maybe remove the changes to `yarn.lock`, too, because they're conflicting with what's on `master`. Easier to merge this without that.

PR #559, opened 2018-10-23
DApp: Nav + drawer style updates, improved Error / Application messaging

am9u:


packages/utils/src/civilHelpers.ts
```diff
@@ -91,3 +91,16 @@ export enum GovernmentParameters {
   appealFee = "appealFee",
   appealVotePercentage = "appealVotePercentage",
 }
+
+export const supportedNetworks: number[] = [
+  3, // rinkeby
```
nickreynolds: rinkeby is 4 tho..

PR #561, opened 2018-10-24
Deploy censorship resistance to mainnet

ritave:


packages/contracts/truffle-config.js
```diff
@@ -31,6 +39,21 @@ module.exports = {
       network_id: 4,
       gasPrice: "20000000000",
     },
+    ledgerMainnet: {
+      provider: function() {
+        var ledgerProvider = require("@joincivil/dev-utils").ledgerProvider;
+        var infura_key = process.env.INFURA_KEY;
+        var accountId = +process.env.LEDGER_ACCOUNT_ID;
```
walfly: is this `+` right?

nickreynolds:
I'm not sure we need the ledger provider stuff, but I guess it doesn't hurt

packages/contracts/truffle-config.js
walfly: @ritave

jorgelo: @walfly, it looks correct, `accountId` needs to be a number and this is a tricky way to coerce.

walfly:
Can't we just use Number class

PR #563, opened 2018-10-24
Fix for when no window.ethereum.isEnabled

tobek:


tobek:
@walfly happens to me right now, with metamask v4.16.0. Wallet onboarding in newsroom manager and post panel are broken for me as a result.

PR #565, opened 2018-10-25
redux refactor

nickreynolds:
- in progress redux refactor

### DIFF
--- a/packages/dapp/src/components/listinglist/ListingListItem.tsx
+++ b/packages/dapp/src/components/listinglist/ListingListItem.tsx
@@ -147,8 +147,10 @@ const mapStateToProps = (
   const { content } = state.networkDependent;
   let charter;
   if (ownProps.newsroom && ownProps.newsroom.data.charterHeader) {
-    charter = content.get(ownProps.newsroom.data.charterHeader);
+    console.log("ownProps.newsroom.data.charterHeader: ", ownProps.newsroom.data.charterHeader);
+    charter = content.get(ownProps.newsroom.data.charterHeader.uri);
   }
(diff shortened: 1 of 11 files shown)


packages/dapp/src/components/listinglist/ListingListItem.tsx
```diff
@@ -147,8 +147,9 @@ const mapStateToProps = (
   const { content } = state.networkDependent;
   let charter;
   if (ownProps.newsroom && ownProps.newsroom.data.charterHeader) {
-    charter = content.get(ownProps.newsroom.data.charterHeader);
+    charter = content.get(ownProps.newsroom.data.charterHeader.uri);
   }
+  console.log("charter: ", charter);
```
am9u: Extra `console.log` statement?

packages/dapp/src/redux/reducers/ui.ts
```diff
@@ -11,7 +11,7 @@ export function ui(state: Map<string, any> = Map<string, any>(), action: AnyActi
   }
 }
 
-export function useGraphQL(state: boolean = true, action: AnyAction): boolean {
+export function useGraphQL(state: boolean = false, action: AnyAction): boolean {
```
am9u: Minor suggestion, but I was thinking that we might want a dapp `constants.ts` file that stores stuff like the default value for `useGraphQL`, so they're easier to find.

nickreynolds: good idea but I'll hold off for now. long term we'll probably want useGraphQL to be in local storage anyways?

PR #575, opened 2018-10-30
Better gasprice on transactions

walfly:


packages/core/src/templates/contract/partials/tx.handlebars
```diff
@@ -22,6 +22,10 @@ public {{this.name}} = {
         {{> params inputs=inputs}}
         txOptions);
     }
+    if (!isDefined(txOptions.gasPrice)) {
+      const getGasPrice = promisify<BigNumber>(self.instance._eth.getGasPrice, self.instance._eth);
+      txOptions.gasPrice = await getGasPrice();
```
nickreynolds: could you add a debug in here that logs the new gas price

PR #590, opened 2018-11-08
no metamask throws an error [ch2669]

walfly:


packages/newsroom-manager/src/Newsroom.tsx
```diff
@@ -171,9 +171,12 @@ class NewsroomComponent extends React.Component<NewsroomProps & DispatchProp<any
     }
   }
 
-  public render(): JSX.Element {
+  public renderManager(): JSX.Element | null {
+    if (!hasInjectedProvider()) {
+      return null;
```
nickreynolds: is there an error message somewhere else that tells user to install metamask?

walfly:
Yeah there is

PR #636, opened 2018-11-26
Upgrade npm-run-all to remove malicious event-stream dependency

kimpers:
As pointed out int #630 event-stream 3.3.6 has malicious code injected that tries to attack bitcoin wallet software. See dominictarr/event-stream/issues/116

This PR upgrades npm-run-all to the latest version which does not depend on this malicious module


dankins:
@kimpers thanks so much for your contribution! we were just sorting this out and you beat us to it!

PR #639, opened 2018-11-27
make the creator of the newsroom also an editor becuase most transact…

walfly:
…ions are half the gas if theres no multisig involved


packages/contracts/contracts/newsroom/NewsroomFactory.sol
```diff
@@ -26,6 +26,7 @@ contract NewsroomFactory is Factory {
   {
     address wallet = multisigFactory.create(initialOwners, initialRequired);
     newsroom = new Newsroom(name, charterUri, charterHash);
+    newsroom.addRole(msg.sender, "editor");
```
nickreynolds: only thing that's a little weird is passing in "editor" instead of using a constant or something

walfly: the constant is set to private on the newsroom class.  but easy to do it the other way by changing it

walfly: I'll change it if you think thats better

nickreynolds: or could add a function in Newsroom `addEditor` that uses its private constant there

walfly: @nickreynolds updated

PR #644, opened 2018-11-29
Show newsroom logo in listing lists

tobek:


packages/components/src/ListingSummary/NewsroomInfo.tsx
```diff
@@ -1,13 +1,18 @@
...
+  StyledListingSummaryNewsroomName,
+} from "./styledComponents";
 import DepositOrStakeAmount from "./DepositOrStakeAmount";
 import NewsroomTagline from "./NewsroomTagline";
 
 const NewsroomInfo: React.SFC<ListingSummaryComponentProps> = props => {
   return (
     <StyledListingSummaryTop>
-      <NewsroomIcon />
+      <NewsroomIcon>{props.logoURL && <NewsroomLogo src={props.logoURL} />}</NewsroomIcon>
```
am9u: Should we have a placeholder logo?

tobek: Good call I'll ask Nguyet - as it is currently, `<NewsroomIcon>` draws a blank grey box, so spacing is the same and it'll look okayish next to other logos, but yeah a default would be good.

Issue #634, opened 2018-11-26
View should auto scroll to the top of the window when a user navigates to a new view 


am9u:
Right now the window stays at the scroll top that the user scrolled to, even when a new view is loaded

am9u:
https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/scroll-restoration.md

PR #651, opened 2018-11-30
Scroll user to top of window on navigation change, when applicable

am9u:
Add helper component to scroll the user back to the top of the window when they navigate to views that need that behavior

Implemented per React Router's documentation suggestion:
https://reacttraining.com/react-router/web/guides/scroll-restoration

Addresses Issue #634


am9u:
Good catch. I just added the utility component to Submit Challenge and Request Appeal as well.

PR #652, opened 2018-11-30
Fixes for manager so that it works standalone in dapp

tobek:


packages/core/src/civil.ts
```diff
@@ -225,11 +225,11 @@ export class Civil {
...
    * Stores content on a content provider (defaults to IPFS)
    * @param content The the data that you want to store, in the future, probably a JSON
-   * @returns A URI that points to the content
+   * @param options Options to be passed to the provider
+   * @returns StorageHeader with info about published content
    */
-  public async publishContent(content: string): Promise<Uri> {
-    const { uri } = await this.contentProvider.put(content);
-    return uri;
+  public async publishContent(content: string, options: object): Promise<StorageHeader> {
```
tobek: This isn't used anywhere in our codebase so should be safe to change the function signature, this way probably makes more sense anyway.

packages/dapp/src/components/newsroom/NewsroomManagementV2.tsx
```diff
@@ -12,6 +12,11 @@ export interface NewsroomManagementState {
 }
 
 export default class NewsroomManagement extends React.Component<NewsroomManagementProps, NewsroomManagementState> {
+  constructor(props: NewsroomManagementProps) {
```
walfly: why is this constructor needed?

packages/dapp/src/components/newsroom/NewsroomManagementV2.tsx
```diff
@@ -28,6 +33,7 @@ export default class NewsroomManagement extends React.Component<NewsroomManageme
         address={this.props.match.params.newsroomAddress}
         theme={DEFAULT_BUTTON_THEME}
         metamaskEnabled={this.state.metamaskEnabled}
+        allSteps={true}
```
walfly: what is this for?

packages/dapp/src/components/newsroom/NewsroomManagementV2.tsx
tobek: needed to set `this.state = {}` otherwise cannot read `metamaskEnabled` state of undefined.

packages/dapp/src/components/newsroom/NewsroomManagementV2.tsx
tobek: Somewhat poorly named but couldn't think of a better one and comment describes the prop in the newsroom component - it's for adding sign constitution and full apply to TCR step, so that both can be used for testing in the dapp, while excluding them from wordpress plugin.

packages/core/src/civil.ts
tobek: I changed the return value so that this could be used via the `civil` instance to publish the charter in sign constitution step, as a fallback in case the component instantiating the newsroom component doesn't pass in an ipfs provider. Here's the exact bit of code: https://github.com/joincivil/Civil/commit/c2d8a558fdfcb1cf9e6bc8e54a3faed824c71f93#diff-5a61d225804b30fd4c7b30222d0dc35d

PR #661, opened 2018-12-04
Mobile UI for DApp Registry page

am9u:
First pass at Mobile UI for Registry page. This will be updated with styling for the secondary tab navigation, for the In Progress tab, in a near update.


packages/components/src/Hero/Hero.tsx
```diff
@@ -1,30 +1,11 @@
 import * as React from "react";
 import styled from "styled-components";
```
sruddy: You can remove `styled`

PR #664, opened 2018-12-06
wait for the block to be confirmed before going further

walfly:
This fixes all the annoying errors with callbacks not firing after transactions


walfly:
Adding everyone because it affects all client-side code

PR #683, opened 2018-12-11
Copy updates

tobek:


packages/components/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "scripts": {
     "build": "yarn clean && tsc && mkdir build/images && cp -a src/images/. build/images",
+    "build:watch": "tsc -w",
```
walfly: can your computer handle all these watchers, if set too many up it crashes

tobek: Oh I'm not even sure, I usually just start watching in the one or two packages I'm currently working on, but I have a different version of typescript globally installed so I can't just run `tsc -w` cause it errors, so I figured it would be good to have an easy way to watch with the package's version.

I just tried watching the whole monorepo now and my computer seemed to be ok with it, but... haven't pushed it very hard.

PR #684, opened 2018-12-12
CIVIL-28 #comment corrects duplicate editors issue

walfly:


packages/components/src/WalletOnboarding.tsx
```diff
@@ -262,12 +262,6 @@ export class WalletOnboarding extends React.Component<WalletOnboardingProps, Wal
             <ManagerSectionHeading>Wallet Connected</ManagerSectionHeading>
             <WalletLabel>Your wallet address</WalletLabel>
             <WalletAddress address={this.props.metamaskWalletAddress} />{" "}
-            {this.props.profileUrl && (
-              <WalletAction>
-                {this.state.justSaved && "Saved. "}
```
tobek: Should we removed this "Saved" notice too? Added because in one user test a while back they weren't sure if it did anything. Just removing the Open Profile link probably suffices.

walfly: i think we want to let them know it was saved

PR #685, opened 2018-12-12
CIVIL-39 fix checkbox spacing

walfly:


walfly:
fixes a bug where in wordpress a different boxsizing gets inherited

PR #705, opened 2018-12-20
Transaction Status modal updates for DApp and components library

am9u:
This PR includes:

- `icons` stories for Storybook, so icons are previewable
- Updated designs, including the correct success icons, for transaction progress modals
- "Transactions are not yet available on mobile" modal, that is enabled w/ some responsive media query


packages/components/src/icons/SubmitChallengeSuccessIcon.tsx
```diff
@@ -0,0 +1,129 @@
...
+                  id="Shape"
+                />
+                <rect id="Rectangle-path" x="31" y="5" width="22" height="2" />
+                <rect id="Rectangle-path" x="31" y="18" width="7" height="2" />
```
sruddy: You should strip ids out of svgs

packages/components/src/icons/CommitVoteSuccessIcon.tsx
```diff
@@ -0,0 +1,43 @@
...
+  height?: number;
+  width?: number;
+}
+
+export const CommitVoteSuccessIcon: React.SFC<CommitVoteSuccessIconProps> = props => {
+  const width = (props.width || 100).toString();
+  const height = (props.height || 100).toString();
+
+  return (
+    <svg xmlns="http://www.w3.org/2000/svg" width={width} height={height} viewBox={`0 0 ${width} ${height}`}>
```
sruddy: @am9u one more thing I just noticed the height and width in the viewBox should stay the original size. Changing this will change the size of the box but not the actual icon so it would either cut it off or have white space.

am9u: Ah yeah, that's a good point. These probably won't need to be resized very often and the effort to make them resizeable may not be worth it right now. I'll make those changes.

PR #708, opened 2018-12-27
contract updates deployments

nickreynolds:
- updating contracts
- unit tests
- updating core/dapp with necessary changes


nickreynolds:
closing without merging due to failing unit test that exposed a bug in web3.js/trufflev4. Upgrading to truffle v5 would fix but is too risky this close to launch. We'll just have to leave the salt stuff in for claiming rewards which was original plan anyways.

PR #719, opened 2019-01-03
CVLToken + CivilTokenController

dankins:
- replaces EIP20 with CVLToken and token controller
- proof of use still isn't 100% baked, but I will tackle that after this is merged
- doesn't update any of the core artifacts yet, will need some advice on how to handle this


packages/contracts/contracts/token/ERC1404/ERC1404.sol
```diff
@@ -0,0 +1,18 @@
+pragma solidity ^0.4.24;
+import "../../zeppelin-solidity/token/ERC20/ERC20.sol";
```
nickreynolds: is this import necessary?

packages/contracts/contracts/token/Managed.sol
```diff
@@ -0,0 +1,35 @@
...
+
+    function checkManagerStatus (address managerAddress) public view returns (bool) {
+        return managers[managerAddress];
+    }
+
+    function isManager () public view returns (bool) {
+        return checkManagerStatus(msg.sender);
+    }
+
+    function addManager (address managerAddress) public onlyManagerOrOwner {
```
nickreynolds: I still think the `addManager` and `removeManager` functions are kind of dangerous. if one manager is compromised the whole thing is.

packages/contracts/contracts/ux/CreateNewsroomInGroup.sol
```diff
@@ -1,27 +1,26 @@
...
-    groups = _groups;
+    controller = _controller;
   }
 
   function create(string name, string charterUri, bytes32 charterHash, address[] initialOwners, uint initialRequired)
     public
     returns (Newsroom newsroom)
   {
-    require(initialOwners.length > 0);
+    // require(initialOwners.length > 0);
```
nickreynolds: why is this commented out?

packages/contracts/migrations/2_optional_for_test.ts
```diff
@@ -18,11 +21,11 @@ module.exports = (deployer: any, network: string, accounts: string[]) => {
   const decimals = "18";
 
   async function giveTokensTo(addresses: string[], originalCount: number): Promise<boolean> {
-    const token = await Token.deployed();
     const user = addresses[0];
     let allocation;
     allocation = 50000000000000000000000;
     console.log("give " + allocation + " tokens to: " + user);
+    const token = await Token.deployed();
```
nickreynolds: not sure why this moved. guess it doesn't matter though.

packages/contracts/test/token/CVLToken.ts
```diff
@@ -0,0 +1,75 @@
...
+    code = await token.detectTransferRestriction.call(accounts[1], accounts[5], 1);
+    expect(code.toNumber()).to.eql(1);
+  });
+  it("should prevent others from changing token controller", async () => {
+    return expect(
+      token.changeController(civilTokenController.address, { from: accounts[5] }),
+    ).to.eventually.be.rejected();
+  });
+
+  it("should fail transfers when using the civil token controller", async () => {
```
nickreynolds: add "should succeed transfers when using the no op token controller"

PR #733, opened 2019-01-04
Case fix for "Parameterizer"

jorgelo:
* Fixes a bug where there are two Parameterizer directories ("Parameterizer" and "parameterizer").


jorgelo:
@nickreynolds Good point. I'm thinking all someone needs to do is remove the lowercase directory from their machine (if it exists like it did on mine) and checkout. We should test this though. Want to be the guinea pig?

nickreynolds:
just checked out your branch and immediately I have:

`deleted:    src/components/Parameterizer/Proposal.tsx`

in git. ;(

Issue #631, opened 2018-11-26
Dashboard doesn't display challenges / listings properly when GraphQL is enabled


am9u:
My Activity/Dashboard needs to hydrate its views from GraphQL when it is enabled

nickreynolds:
I actually don't think it needs to hydrate via GraphQL. Right now the graphql queries won't work for a user's historic actions. So we still use the web3 way of getting data for user info, there must just be some small issue preventing this from working correctly when graphql is selected.

PR #737, opened 2019-01-07
load listing data for dashboard activity items

nickreynolds:
- so they show up properly when loading via graphql
- kind of a hack until we get better dashboard support via graphql


nickreynolds:
fixes: https://github.com/joincivil/Civil/issues/631

Issue #513, opened 2018-10-10
Submit Challenge / Request Appeal forms are missing required validation


am9u:
If the user leaves fields blank in these views, they are still able to click the transaction button and the transaction silently fails in the UI.

So we need to add some validation that disables the transaction button and message the user about missing fields.

PR #739, opened 2019-01-07
Add validation messaging to Submit Challenge Statement / Request Appeal forms

am9u:
Issue: #513


packages/components/src/UserStatement/RequestAppealStatement.tsx
```diff
@@ -178,6 +177,13 @@ export class RequestAppealStatement extends React.Component<RequestAppealStateme
     return !summaryValue || !summaryValue.length || !citeConstitution.innerText.length || !details.innerText.length;
   };
 
+  private renderHelperMessage = (): JSX.Element => {
+    if (this.isFormInvalid()) {
```
nickreynolds: does this actually block the user from moving forward? I think it'd be good if the button was disabled EITHER if the field is empty OR invalid (but empty wouldn't have error message?)

am9u: The only issue is since the fields are required, if a user leaves a field empty, there is no messaging that informs the user that they're missing a required field (other than a tiny bit of text by the input field).

Maybe a better solution would be to never disable the button and add the validation as a preTransactions method... or add support for a separate validator fn prop for the Transaction button, which would then trigger the display of messaging.

PR #745, opened 2019-01-08
Fix text for Request Appeal form

am9u:
Addresses #519 

Additionally, this update:

- fixes the rendering of Appeal Statement and Appeal Challenge Statement in the DApp
- makes the 'Discussions' tab the initially active tab, if a challenge is active for the listing


am9u:
@nickreynolds are you suggesting the copy be changed for that input field? or that the field (citing the constitution) be removed from the form?

I just found the copy for that field that says:

```
Please cite sections of the Civil Constitution that you believe support your position on why the vote should be overturned. (required)
```

Is that what you had in mind?

PR #747, opened 2019-01-09
Token account landing page

sruddy:


sruddy:
Note: there's a ticket to add links and props once content and design are finalized

Issue #753, opened 2019-01-14
Newsroom Under Challenge Not Loading via web3


nickreynolds:
- if you load listings via web3 (not graphql) and then go to listing under challenge, page crashes with error rendering challenge statement
- if you load listing under challenge via graphql then turn off graphql, data goes away and page is basically blank

PR #754, opened 2019-01-14
Fix Challenge Statement Rendering

nickreynolds:
- getting challenge statements to load properly via graphql and web3


nickreynolds:
should fix https://github.com/joincivil/Civil/issues/753

Issue #743, opened 2019-01-08
Owner Listing Actions Not Converting Token Amounts Correctly


nickreynolds:
typing "1" should deposit 1 CVL. Right now it deposits .00000000000000001 CVL.

PR #785, opened 2019-01-17
multiply owner action token amounts by 1e18

nickreynolds:


nickreynolds:
fixes https://github.com/joincivil/Civil/issues/743

PR #797, opened 2019-01-21
Quiz modal

sruddy:
- quiz welcome screens
- quiz landing
- launch from token page


packages/components/src/icons/index.ts
```diff
@@ -10,13 +10,15 @@ export * from "./CvlToken";
 export * from "./ClockIcon";
 export * from "./CloseXIcon";
 export * from "./CommitVoteSuccessIcon";
+export * from "./DisclosureArrowIcon";
```
am9u: Should we add these 3 new icons to the icon stories for Storybook?

PR #803, opened 2019-01-22
a few tiny fixes

nickreynolds:
fixes for challenging granted appeal, voting on granted appeal challenge, event timestamps


packages/dapp/src/components/listing/AppealChallengeCommitVote.tsx
```diff
@@ -305,11 +305,12 @@ class AppealChallengeCommitVote extends React.Component<
 
   private commitVoteOnChallenge = async (): Promise<TwoStepEthTransaction<any>> => {
     const voteOption: BigNumber = new BigNumber(this.state.voteOption as string);
-    const saltStr = fetchSalt(this.props.challengeID, this.props.user);
+    const pollID = this.props.appealChallengeID ? this.props.appealChallengeID : this.props.challengeID;
```
am9u: I think this could probably just be `this.props.appealChallengeID`, since this view is specific to appeal challenges. But this works as well.

PR #811, opened 2019-01-24
Authenticated routes 

jorgelo:
* Adds a helper that redirects if a user is not authenticated.
* Also has a param that reverses this functionality and only allows if they are *not* authenticated.
* Maybe a bit of overkill, but it calls the `currentUser` graphql query and redirects on (any) error.

### DIFF
--- a/packages/components/src/AuthenticatedRoute.tsx
+++ b/packages/components/src/AuthenticatedRoute.tsx
@@ -34,8 +34,6 @@ export const AuthenticatedRoute = ({
 }: AuthenticatedRouteProps) => {
   const auth = getApolloSession();
 
-  console.log("AuthenticatedRoute:", { auth });
-
   const hasAuthToken = !!(auth && auth.token);
 
   if (onlyAllowUnauthenicated === hasAuthToken) {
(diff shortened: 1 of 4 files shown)


packages/components/src/AuthenticatedRoute.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+const userQuery = gql`
+  query {
+    currentUser {
+      uid
+      email
+      ethAddress
+      onfidoApplicantId
+      onfidoCheckID
+      kycStatus
```
sruddy: We can probably remove `onfidoApplicantId` `onfidoCheckID` `kycStatus` since we no longer need kyc

packages/dapp/src/components/Account/UserInfo.tsx
```diff
@@ -13,7 +13,6 @@ const userQuery = gql`
       kycStatus
```
sruddy: Same here: remove `onfidoApplicantId` `onfidoCheckID` `kycStatus` since we no longer need kyc

packages/components/src/AuthenticatedRoute.tsx
jorgelo: Removed.

packages/dapp/src/components/Account/UserInfo.tsx
jorgelo: Removed.

PR #825, opened 2019-01-28
*Rough* fix for appeal challenges my activity

nickreynolds:
- adds appeal challenges to rescue and claim rewards for to my activity
- these do not necessarily render properly, the whole my activity section will be overhauled this sprint, but it does unblock users who want to transfer tokens out of voting contract

### DIFF
--- a/packages/dapp/src/components/Dashboard/ActivityList.tsx
+++ b/packages/dapp/src/components/Dashboard/ActivityList.tsx
@@ -2,10 +2,15 @@ import * as React from "react";
 import { Set } from "immutable";
 import BigNumber from "bignumber.js";
 import { ChallengeActivityListItem, ActivityListItem, ResolvedChallengeActivityListItem } from "./ActivityListItem";
+import {
+  ResolvedAppealChallengeActivityListItem,
+  // AppealChallengeActivityListItem,
+} from "./ActivityListAppealChallengeItem";
 
 export interface ActivityListOwnProps {
   listings?: Set<string>;
   challenges?: Set<string>;
+  appealChallenges?: Set<string>;
   resolvedChallenges?: boolean;
   toggleChallengeSelect?(challengeID: string, isSelected: boolean, salt: BigNumber): void;
 }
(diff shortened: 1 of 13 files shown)


packages/dapp/src/components/Dashboard/ActivityList.tsx
```diff
@@ -37,12 +42,38 @@ class ActivityList extends React.Component<ActivityListOwnProps> {
...
           this.props.challenges.map(c => {
             index++;
             return <ChallengeActivityListItem key={c} challengeID={c!} even={index % 2 === 0} />;
           })}
+        {!this.props.resolvedChallenges &&
+          this.props.appealChallenges &&
+          this.props.appealChallenges.map(c => {
+            index++;
+            return (
+              <ResolvedAppealChallengeActivityListItem
```
am9u: Should this be `AppealChallengeActivityListItem`?

packages/dapp/src/components/Dashboard/ActivityList.tsx
```diff
@@ -2,10 +2,15 @@ import * as React from "react";
 import { Set } from "immutable";
 import BigNumber from "bignumber.js";
 import { ChallengeActivityListItem, ActivityListItem, ResolvedChallengeActivityListItem } from "./ActivityListItem";
+import {
+  ResolvedAppealChallengeActivityListItem,
+  // AppealChallengeActivityListItem,
```
am9u: Is this import intentionally commented out?

packages/dapp/src/components/Dashboard/ActivityList.tsx
nickreynolds: as far as I can tell the only difference between the 2 components is that one supports the check-box. I think that we will end up removing the "Resolved" version of both components during this sprint's refactor but wanted to leave it in for now to sorta line up with the non-appeal-challenge versions

packages/dapp/src/components/Dashboard/ActivityList.tsx
nickreynolds: yeah just to remind us to take a look at it during refactor

PR #835, opened 2019-01-29
Update GraphQL URI Based on Network

nickreynolds:
- save network in local storage, update network in apollo client based on local storage

### DIFF
--- a/packages/utils/src/apolloClient.ts
+++ b/packages/utils/src/apolloClient.ts
@@ -13,6 +13,7 @@ export interface AuthLoginResponse {
 let client: ApolloClient<NormalizedCacheObject>;
 
 const SESSION_KEY = "apollo_session";
+const NETWORK_KEY = "nework";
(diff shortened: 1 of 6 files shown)


jorgelo:
@nickreynolds Since the session is attached to the endpoint, when the network changes, we would need to store a separate session for each network/graphql target (or just delete it and have them logout).

Right now, if the user is logged in, and they change the network, all graphql queries will fail since the token is invalid on that endpoint.

packages/utils/src/apolloClient.ts
```diff
@@ -13,17 +13,34 @@ export interface AuthLoginResponse {
 let client: ApolloClient<NormalizedCacheObject>;
 
 const SESSION_KEY = "apollo_session";
+const NETWORK_KEY = "nework";
```
jorgelo: Possibly a typo :)

PR #838, opened 2019-01-30
change createNewsroom URL to create-newsroom to match standard

nickreynolds:
fixes https://github.com/joincivil/Civil/issues/837


nickreynolds:
fixed

Issue #859, opened 2019-02-05
CVL available to claim is inaccurate


nickreynolds:
- said 16.67 CVL but was actually 50
<img width="1223" alt="screen shot 2019-02-04 at 6 48 59 pm" src="https://user-images.githubusercontent.com/1156755/52245531-05501400-28b0-11e9-8aef-f57e8dbbc103.png">
<img width="1276" alt="screen shot 2019-02-04 at 7 06 10 pm" src="https://user-images.githubusercontent.com/1156755/52245537-07b26e00-28b0-11e9-8e02-47f1be60508b.png">

PR #861, opened 2019-02-05
Fix Unclaimed Reward Amount

nickreynolds:
get unclaimed reward amount directly from contract instead of doing complicated things


nickreynolds:
fixes https://github.com/joincivil/Civil/issues/859

PR #865, opened 2019-02-05
Storefront Airswap component $$$

sruddy:


packages/components/src/Airswap/AirswapBuyCVL.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from "react";
+import makeAsyncScriptLoader from "react-async-script";
+import { Button } from "../Button";
+
+const AIRSWAP_URL = "https://cdn.airswap.io/gallery/airswap-trader.js";
```
am9u: Hmm, we should maybe put these configs into `@joincivil/utils` along with other config / network derived values.

packages/dapp/src/components/Tokens/Tokens.tsx
```diff
@@ -1,18 +1,48 @@
...
       </Helmet>
       <ScrollToTopOnMount />
-      <UserTokenAccount />
+      <UserTokenAccount
+        userLoggedIn={true}
+        userTutorialComplete={false}
+        userAccount={props.userAccount}
+        supportEmailAddress={"support@civil.co"}
+        faqUrl={"https://cvlconsensys.zendesk.com/hc/en-us"}
+        foundationAddress={"0xf1176B0aeb7914B5472B61c97A4CF0E0bcacB579"}
```
am9u: Is this the appellate multisig? Should this come from some utils thing that specifies RInkeby vs Mainnet vs Ganache so it can be tested in diff environments w/o affecting mainnet?

packages/components/src/Airswap/AirswapBuyCVL.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    // @ts-ignore
+    window.AirSwap.Trader.render(
+      {
+        mode: "buy",
+        env: "sandbox",
+        token: rinkeby,
+        address: buyFromAddress,
+        onComplete: (transactionId: string) => {
+          console.info("Trade complete. Thank you, come again.", transactionId);
```
am9u: Are these `console` statements placeholders for future messaging to the user when these methods fire?

sruddy: Yes, this functionality is coming soon. We haven't decided what is going to happen yet.

packages/dapp/src/components/Tokens/Tokens.tsx
sruddy: We currently use the same address in all environments and Airswap switches between rinkeby and mainnet I will add those to utils though.

packages/components/src/Airswap/AirswapBuyCVL.tsx
sruddy: I will add this adding a TODO for now

PR #866, opened 2019-02-06
Rejected Listing Item Fixes

nickreynolds:
several fixes for rejected listing items: get appeal challenge to show properly, get appeal decision to be correct, show if appeal was requested


packages/components/src/ListingDetailPhaseCard/AppealChallengeResolveCard.tsx
```diff
@@ -37,6 +29,13 @@ export type AppealChallengeResolveCardProps = ListingDetailPhaseCardComponentPro
   AppealChallengeResultsProps;
 
 export const AppealChallengeResolveCard: React.SFC<AppealChallengeResolveCardProps> = props => {
+  const showAppealChallenge =
+    props.appealChallengeTotalVotes &&
+    props.appealChallengeVotesFor &&
+    props.appealChallengeVotesAgainst &&
+    props.appealChallengePercentFor &&
+    props.appealChallengePercentAgainst &&
+    props.didAppealChallengeSucceed;
```
am9u: I think `didAppealChallengeSucceed` shouldn't be part of this check, as we want the results to probably show regardless of the result of the vote, right?

PR #867, opened 2019-02-06
Move auth into newsroom-signup + flow cleanup + apollofication of ETH auth flow + fixes

tobek:
Note if you want to test this: the link in the magic link email isn't set up for newsroom signup yet, so just take the token from the email and manually go to `/signupNewsroom?token=[token]` to verify the token and log in.


packages/components/src/Account/Auth/EthAuth.tsx
```diff
@@ -113,10 +153,24 @@ export class AccountEthAuth extends React.Component<AccountEthAuthProps, Account
         },
         postTransaction: async (sig: EthSignedMessage): Promise<void> => {
           try {
-            await userSetEthAddress(sig);
-            this.setState({ isWaitingSignatureOpen: false });
-            if (this.props.onAuthenticated) {
-              this.props.onAuthenticated(sig.signer);
+            delete sig.rawMessage; // gql endpoint doesn't want this and errors out
```
walfly: why delete instead of `= null`;

packages/components/src/Account/Auth/EthAuth.tsx
```diff
@@ -113,10 +153,24 @@ export class AccountEthAuth extends React.Component<AccountEthAuthProps, Account
...
+              },
+            });
+
+            if (res && res.data && res.data.userSetEthAddress) {
+              if (this.props.onAuthenticated) {
+                this.props.onAuthenticated(sig.signer);
+              }
+              if (this._isMounted) {
+                // A bit of an antipattern, but cancelling async/await is hard
+                this.setState({ isWaitingSignatureOpen: false });
```
walfly: theres definitely a better way to do this.  Why would this code ever run when the component isn't mounted.  Should you just move these handlers up a level in the tree?

tobek: I'm not sure how to avoid this. This component updates `currentUser.ethAddress` on the server, and the existence of `currentUser.ethAddress` is used by parent component to determine whether or not this component needs to be displayed (https://github.com/joincivil/Civil/blob/tobek/newsroom-auth/packages/components/src/WalletOnboarding.tsx#L267-L272). But it looks like before the update mutation actually returns here, the Apollo cache updates the query and this component gets unmounted.

This component is reused in multiple places (storefront and newsroom signup, maybe elsewhere in the future) whenever ETH auth is needed, so this logic can't be moved up a level in the tree without duplicate code, or creating some wrapper component that's always mounted by its parent but returns null if it's not needed.

Can't think of how to get around this without a bunch of refactoring that would probably make this code more complicated.

packages/components/src/Account/Auth/EthAuth.tsx
tobek: `sig` is of type `EthSignedMessage` which expects `rawMessage` to be a string, so setting it to undefined or null is a type error, but (for some reason) deleting it doesn't. Other option is to typecast in a couple places or tslint ignore that line. Any reason not to use `delete`?

PR #873, opened 2019-02-06
Connect storefront to backend and add wallet component.

jorgelo:
* Adds logic to redirect to connect your wallet.
* Includes Toby's add wallet component.
* Covers tickets CIVIL-307 and CIVIL-310
* Adds a nice little helper called LoadUser that returns the user if there is one.


packages/dapp/src/components/Account/index.tsx
```diff
@@ -92,6 +111,17 @@ export default class AccountRouter extends React.Component<RouteComponentProps>
...
   };
 
+  public handleOnAddWallet = (): void => {
+    const { history } = this.props;
+
+    history.push({
+      pathname: `/tokens`,
+      state: {},
+    });
+    // TODO(jorgelo): Same as above, there should be a better way to force a refetch.
```
tobek: Check this again after we merge my PR https://github.com/joincivil/Civil/pull/867 with updated AccountEthAuth component - the `update` property passed to `<Mutation>` updates the cached value for the `currentUser { ethAddress }` query, which I damn hope also updates the cached value for the larger `currentUser` query you use, so it should update state without refresh.

packages/dapp/src/components/Account/LoadUser.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const hasAuthToken = auth && auth.token;
+
+  if (!hasAuthToken) {
+    return <>{renderFunction({ user: null, loading: false })}</>;
+  }
+
+  return (
+    <Query query={userQuery}>
+      {({ loading, error, data }) => {
+        if (loading || error) {
```
tobek: If you haven't already you should double check that this handles the logged-out state (which will cause this request to error) in the way you expect - unless you're only using this component in routes that require authentication?

packages/components/src/Tokens/Tokens.tsx
```diff
@@ -27,14 +31,18 @@ export class UserTokenAccount extends React.Component<{}, UserTokenAccountStates
...
+
     return (
       <TokenAccountOuter>
         <TokenAccountInner>
           <UserTokenAccountHeader />
           <FlexColumns>
             <FlexColumnsPrimary>
               <FlexColumnsPrimaryModule>
-                <UserTokenAccountSignup step={"completed"} />
+                <UserTokenAccountSignup step={accountSignupStep} />
```
sruddy: I updated this flow a little bit in a new branch so we're going to have some merge conflicts to work out. Should be easy fixes but there are a few things we should discuss. Same with `TokensAccountSignup.tsx`

PR #874, opened 2019-02-07
Updates and bug fixes for Claiming Rewards and Rescuing Tokens in the Dashboard

am9u:
- Update both views to show proper token amounts for Appeal Challenges,
by getting the proper appeal challenge data for a `appealChallengeID`
- Remove selectors for getting unclaimed amounts for individual
challenges, as that value is now read directly from event logs / helpers
in `@joincivil/core`
- Iterate on UI to match latest mocks for items in those tabs

### DIFF
--- a/packages/core/src/contracts/tcr/civilTCR.ts
+++ b/packages/core/src/contracts/tcr/civilTCR.ts
@@ -555,6 +555,7 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
     let salt;
     let numTokens;
     let choice;
+    let voterReward;
     const [, , resolved] = await this.instance.challenges.callAsync(appealChallengeID);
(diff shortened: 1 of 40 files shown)


packages/core/src/contracts/tcr/civilTCR.ts
```diff
@@ -590,6 +591,10 @@ export class CivilTCR extends BaseWrapper<CivilTCRContract> {
       didCollectAmount = await this.getRewardClaimed(appealChallengeID, user);
     }
 
+    if (isVoterWinner && !didUserCollect) {
+      voterReward = await this.voterReward(appealChallengeID, salt as BigNumber, user);
```
nickreynolds: this is weird... didn't I add this already?

am9u: This is for appeal user challenge data!

PR #877, opened 2019-02-08
Create data wrapper to integrate newsroom-signup with API

tobek:
This creates a `DataWrapper` component that gets the required queries/mutations from GQL and passes them into `Newsroom` - it's a bit roundabout/awkward to use both Redux and Apollo for saving/loading data, but we're only using a small handful of things from Apollo and it works to keep them separately. Feedback welcome.

PR against `tobek/newsroom-auth` branch because it relies on code from that PR #867 which still needs review.


packages/newsroom-signup/src/DataWrapper.tsx
```diff
@@ -0,0 +1,108 @@
...
+            </Query>
+          );
+        }}
+      </Query>
+    );
+  }
+
+  private saveCharterFuncFromMutation(mutation: MutationFunc): (charter: Partial<CharterData>) => Promise<void> {
+    return async (charter: Partial<CharterData>) => {
+      await mutation({
```
walfly: shouldn't this return?

packages/newsroom-signup/src/Newsroom.tsx
```diff
@@ -197,12 +185,14 @@ class NewsroomComponent extends React.Component<NewsroomProps & DispatchProp<any
...
 
-    this.props.dispatch!(addPersistCharter(this.persistCharter));
+    this.props.dispatch!(addPersistCharter(this.debouncedPersistCharter));
+    this.initCharter();
 
-    if (this.props.address && this.props.civil) {
-      await this.hydrateNewsroom(this.props.address);
-    }
     if (this.props.civil) {
+      if (this.props.address) {
```
walfly: good catch

packages/newsroom-signup/src/DataWrapper.tsx
tobek: Yep good call. Newsroom component isn't using return value of `persistCharter` at the moment but makes sense to pass it through anyway.

PR #883, opened 2019-02-11
Tutorial bugs and content update

sruddy:


packages/components/src/TokenTutorial/TokenTutorialQuiz.tsx
```diff
@@ -60,21 +65,25 @@ export class TokenTutorialQuiz extends React.Component<TokenTutorialQuizProps, T
...
+                answer={TutorialContent[topicIdx].questions[slideIdx].answer}
+                onClickPrev={() => this.prev()}
+                onClickNext={() => this.next()}
+                activeSlide={slideIdx + 1}
+                totalSlides={TutorialContent[topicIdx].questions.length}
+              />
+            </TutorialContain>
+          );
+        } else {
+          this.resetQuestion();
```
jorgelo: Changing the state on render is kind of unpredictable.  Are you trying to force a re-render? That can be done with `forceUpdate`: https://reactjs.org/docs/react-component.html#forceupdate . 

That being said, if this works, cool, leave it be.

sruddy: @jorgelo I’m not a fan of this but `forceUpdate` doesn't do exactly what I need. I’m going to test this and try looking into a better fix.

PR #886, opened 2019-02-11
Newsroom Withdrawls

nickreynolds:
- add support for newsroom to accept/withdraw ETH and to withdraw ERC20s sent to it. tests.


dankins:
How will this be deployed? Seems like we would need to do some hackery with migrations and then patching the artifacts manually? How will we manage Newsroom contract updates in the future? Is the "more correct" answer to have Newsroom contract be upgradable?

nickreynolds:
so we would just replace the existing NewsroomFactory and all existing newsrooms would create new ones and forget about their old ones.

upgradability is not the solution, imo. Future upgrades could do things like point back to this newsroom though.

nickreynolds:
not doing this right now

Issue #790, opened 2019-01-18
Listing History Items Have Alignment Issue


nickreynolds:
- some history event items are being pushed to the right for some reason

<img width="716" alt="screen shot 2019-01-17 at 7 19 58 pm" src="https://user-images.githubusercontent.com/1156755/51357439-f59d9680-1a8c-11e9-9c85-03a3f067db81.png">
<img width="738" alt="screen shot 2019-01-17 at 7 19 35 pm" src="https://user-images.githubusercontent.com/1156755/51357440-f59d9680-1a8c-11e9-9882-bf444655b6ea.png">

PR #894, opened 2019-02-12
Fix History Event Alignment

nickreynolds:
- make sure event date is always 148px wide so it doesn't get squished when rest of history event is bigger
- if there's a better way to do this, let me know


nickreynolds:
fixes https://github.com/joincivil/Civil/issues/790

Issue #885, opened 2019-02-11
Support transparent and non-square newsroom logos


tobek:
1. Don't crop non-square logos (on listings list and newsroom detail page header)
2. Logo should have white bg (on newsroom detail page header)

See Sludge on https://registry.civil.co/

PR #899, opened 2019-02-12
Newsroom Listing Logo Fixes

nickreynolds:
- use contain instead of cover for newsroom logo object-git
- set background to white


nickreynolds:
fixes https://github.com/joincivil/Civil/issues/885

PR #900, opened 2019-02-12
Connects Quiz to Backend

jorgelo:
* Connects quiz to back end (CIVIL-249)
* Saves and reads 
* Fixes a dumb redirect bug.

### DIFF
--- a/packages/components/src/TokenTutorial/TokenTutorialLanding.tsx
+++ b/packages/components/src/TokenTutorial/TokenTutorialLanding.tsx
@@ -74,7 +74,6 @@ export class TokenTutorialLanding extends React.Component<TokenTutorialLandingPr
         </TutorialSkipSection>
 
         {TutorialContent.map((topic, idx) => {
-          console.log("???", this.props.quizPayload, this.props);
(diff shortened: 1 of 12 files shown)


packages/components/src/TokenTutorial/TokenTutorialLanding.tsx
```diff
@@ -70,29 +73,76 @@ export class TokenTutorialLanding extends React.Component<TokenTutorialLandingPr
...
-                <b>0/{topic.questions.length}</b>
-              </TutorialLandingProgressBars>
-            </TopicProgress>
-          </TutorialTopic>
-        ))}
+        {TutorialContent.map((topic, idx) => {
+          const { isComplete, lastSlideIdx } = this.getTopicStatus(this.props.quizPayload, topic);
+
+          // TODO(jorgelo): What do we do when isComplete is true (this means that this topic has been completed)
+          // TODO(jorgelo): lastSlideIdx is the last slide that was completed correctly. Should we jump the user to that last slide?
```
sruddy: @jorgelo I think if you pass in `quizSlide` from the landing page it should work. You would still see the tutorial but then it should take you to the question you left off on.

Issue #880, opened 2019-02-09
Newsroom Passed Without Challenge Showing Vote Results


nickreynolds:
<img width="439" alt="screen shot 2019-02-09 at 1 28 41 pm" src="https://user-images.githubusercontent.com/1156755/52525204-2cbe2c80-2c74-11e9-8000-6dfb77d210e6.png">

PR #902, opened 2019-02-13
Tiny Listing Card Fix

nickreynolds:
- don't show challenge results on ready-to-update newsroom if there is no challenge


nickreynolds:
fixes https://github.com/joincivil/Civil/issues/880

PR #930, opened 2019-02-18
Adding Retry Logic to getIPFSContent

nickreynolds:
- adding retry logic to getIPFSContent since this fails mysteriously a fair amount, but subsequent attempts are likely successful

### DIFF
--- a/packages/dapp/src/helpers/listingEvents.ts
+++ b/packages/dapp/src/helpers/listingEvents.ts
@@ -137,14 +137,27 @@ export async function getNewsroom(dispatch: Dispatch<any>, address: EthAddress):
   }
 }
 
-export async function getIPFSContent(header: StorageHeader, dispatch: Dispatch<any>): Promise<void> {
+async function delay(ms: number): Promise<any> {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+export async function getIPFSContent(
+  header: StorageHeader,
+  dispatch: Dispatch<any>,
+  wait: number = 1000,
+): Promise<void> {
   const civil = getCivil();
   const content = await civil.getContent(header);
   if (content) {
     const parsedContent = JSON.parse(content.toString());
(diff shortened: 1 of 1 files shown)


packages/dapp/src/helpers/listingEvents.ts
```diff
@@ -137,14 +137,27 @@ export async function getNewsroom(dispatch: Dispatch<any>, address: EthAddress):
...
   if (content) {
     const parsedContent = JSON.parse(content.toString());
     dispatch(addContent(header, parsedContent));
   } else {
-    console.error("Missing IPFS content for header:", header);
+    console.warn("Missing IPFS content for header:", header);
+    if (header.uri !== "") {
+      console.warn("Retrying getIPFSContent. wait = " + wait + "ms");
+      await delay(wait);
+      return getIPFSContent(header, dispatch, wait * 2);
```
am9u: Should we have a `maxRetries` or something?

nickreynolds: makes sense, just added some logic so it only tries 6 times max (6 is arbitrary)

packages/dapp/src/helpers/listingEvents.ts
```diff
@@ -137,14 +137,30 @@ export async function getNewsroom(dispatch: Dispatch<any>, address: EthAddress):
...
   if (content) {
     const parsedContent = JSON.parse(content.toString());
     dispatch(addContent(header, parsedContent));
   } else {
-    console.error("Missing IPFS content for header:", header);
+    console.warn("Missing IPFS content for header:", header);
+    if (header.uri !== "" && tries < 6) {
+      console.warn("Retrying getIPFSContent. wait = " + wait + "ms");
+      await delay(wait);
+      return getIPFSContent(header, dispatch, wait * 2, tries + 1);
```
walfly: why double the wait?

PR #938, opened 2019-02-19
Style MetaMask onboarding flow

tobek:
- Forks `<WalletOnboarding>` used in WP plugin to `<WalletOnboardingV2>` in `newsroom-signup` package cause it's really different
- Styles it, luxuriously
- Stops using `ethereum.isEnabled`, just uses `ethereum.enable`

@jorgelo you may want to use `<WalletOnboardingV2>` instead of just `<AccountEthAuth>` alone. If you haven't seen it check out the whole onboarding flow starting at https://consensys.invisionapp.com/share/AVPK9VRRMNX#/336620439_02 and ending at the "Wallet Connected" state. I assume you guys need something like this, except that after hitting Continue on the final "Wallet Connected" state you still want to display a small widget showing you're connected? But the whole flow kind of needs a lot of space, so for newsroom signup we simply hide the everything until you've gone through wallet onboarding: https://github.com/joincivil/Civil/blob/master/packages/newsroom-signup/src/Newsroom.tsx#L335. If you want to do something like this let's talk!

(Note that  `<WalletOnboardingV2>` is in the `newsroom-signup` package right now. You can import it from `dapp` but not from `components` - latter will make a circular dependency and breaks. If you need that I can fix it by moving it into `components`, just needs some refactoring to move a bunch of shared `newsroom-signup` stuff around.)


packages/utils/src/crypto.ts
```diff
@@ -71,3 +71,21 @@ export const networkNames: { [index: string]: string } = {
...
+      return !!(await (window as any).ethereum.enable());
+    } catch (e) {
+      return false;
+    }
+  } else {
+    return true;
+  }
+}
+
+export async function enableEthereum(): Promise<any> {
```
walfly: seems redundant with the other method above

tobek: Yeah true. The only difference is that the above returns a boolean, the below returns a... well MM returns an array of one address. Will replace them both with a single `ethereumEnable` in my next PR.

PR #941, opened 2019-02-19
Copy and UI updates for Listing Detail page and Voting flows

am9u:


packages/components/src/ListingDetailPhaseCard/ChallengePhaseDetail.tsx
```diff
@@ -10,26 +10,35 @@ export const ChallengePhaseDetail: React.StatelessComponent<ChallengePhaseProps>
...
+          </MetaItemLabel>
+          <MetaItemValue>{props.rewardPool}</MetaItemValue>
         </MetaItem>
       </MetaRow>
       <MetaRow>
         <MetaItem>
           <MetaItemLabel>
-            Reward Pool
-            <QuestionToolTip explainerText={<RewardPoolToolTipText />} positionBottom={true} />
+            Challenger Deposit
```
nickreynolds: I believe the Reward Pool is dispPct * deposit, isn't it?

packages/dapp/src/components/listinglist/ListingListItem.tsx
```diff
@@ -142,6 +142,7 @@ export const ListingItemBaseComponent: React.SFC<
   const ListingSummaryItem = props.ListingItemComponent || ListingSummaryComponent;
   const listingViewProps = transformListingSummaryViewProps(props);
 
+  console.log(listingViewProps);
```
nickreynolds: extra log

packages/components/src/ListingDetailPhaseCard/ChallengePhaseDetail.tsx
am9u: You're right. I just updated the tooltip to show the correct dispensationPct from Parameters.

PR #942, opened 2019-02-19
add govt proposal support to dapp

nickreynolds:


packages/dapp/src/helpers/government.ts
```diff
@@ -30,11 +36,99 @@ export async function initializeGovernmentParamSubscription(dispatch: Dispatch<a
   const civil = getCivil();
   const current = await civil.currentBlock();
   const govt = await tcr.getGovernment();
+  const appellate = await govt.getAppellate();
+  console.log("appellate: ", appellate);
```
am9u: extra `console.log`?

PR #955, opened 2019-02-21
Verify Token Bug Fix

jorgelo:
* This is mostly a bug fix for the refresh bug on verify token. The bug happened when a user's metamask network differed from the default (example: Your MM is on mainnet (network 1), but the dapp default is network 4.), it would cause the auth Route components to refresh, querying for a valid token twice. If a user was registering, they would always see a "user already exists with this identifier" error, even though it did work. The root cause was my use of `component=` with a function in React Router instead of `render=`. When using `component` with a function, the function is refreshed every time a state/prop change happens from the parent.
* Some TODO notes to not forget.
* Minor refactoring.


packages/components/src/Account/Auth/AuthStyledComponents.tsx
```diff
@@ -102,8 +102,8 @@ export const AuthFooterTerms = () => (
   <AuthFooterContainer>
     <AuthTextFooter />
     <BenefitsLink>
-      {/* // TODO(jorgelo): Where will this link? */}
-      <Link to={"https://civil.co"} target="_new">
+      {/* // TODO(jorgelo): Confirm this is the final link */}
+      <Link to={"https://civil.co/become-a-member"} target="_new">
```
dankins: maybe move this to `src/helpers/config.ts` (and .env for a default)?

PR #963, opened 2019-02-21
Storefront Metamask Login, Bug fixes, Text Changes.

jorgelo:
* Major feature here is the updated Metamask screens.
* I changed a bit how the token storefront home page buttons look. I put in holder text, I just wanted to show the new logic.
* Minor refactors and text changes.
* Adds some more TODOs as I find them.


packages/components/src/AuthenticatedRoute.tsx
```diff
@@ -26,21 +25,26 @@ const userQuery = gql`
...
-  ethSignupPath,
   ...otherProps
 }: AuthenticatedRouteProps) => {
   const auth = getApolloSession();
 
   const hasAuthToken = !!auth && !!auth.token;
 
-  if (onlyAllowUnauthenticated === hasAuthToken) {
-    return <Redirect to={redirectTo} />;
+  if (onlyAllowUnauthenticated) {
```
dankins: this boolean logic is kinda weird, why not:

if (onlyAllowUnauthenticated && hasAuthToken)
else if (!hasAuthToken)

jorgelo: Noted, I'm going to leave this as is for this PR, but I'll take care of it in a PR for tomorrow.

PR #966, opened 2019-02-22
Support actual signup/login magic links [CIVIL-290]

tobek:


dankins:
can you please squash these commits and update the message before you merge?

tobek:
Certainly the linting and typos and little fixes etc. can be squashed, and rebasing should be preferred over merging, but I don't know about squashing whole PRs - personally I try to make sure each commit is a self contained change and is fully functional and has a meaningful commit message, so that git blame and history are useful (especially for rare but life-saving git bisect), where a large PR squashed to a single "Implement big feature XYZ" commit would be less helpful.

This PR in particular can be squashed without losing anything really, and I'm happy to go with whatever convention we settle on, but I think it'd be great to discuss this bit of git strategy as a team!

PR #968, opened 2019-02-22
Verify Token Page

jorgelo:
* Styled Verify Token Page.
* Took a shot at an error state page.
* More TODOs


dankins:
can you please squash these commits and update the message before you merge?

PR #980, opened 2019-02-25
Storefront bugs

sruddy:
- update storefront after quiz completed
- "Start the Civil Tutorial" should say "Continue" if you already started
- question numbers on progress bar
- fix quiz styles
- drop in new quiz content
- after users click on “continue” skip the 3 welcome screens


packages/components/src/TokenTutorial/TokenTutorial.tsx
```diff
@@ -28,13 +28,26 @@ export class TokenTutorial extends React.Component<TokenTutorialProps, TokenTuto
...
           }
 
           return <TutorialWelcomeScreens handleClose={this.props.handleClose} quizPayload={quizPayload} />;
         }}
       </Query>
     );
   }
+
+  private isQuizStarted = (quizPayload: any) => {
+    if (Object.keys(quizPayload).length > 0) {
```
am9u: Just curious, why not just return `Object.keys(...)`?

sruddy: 👍 I started out doing something else and didn't notice that

PR #985, opened 2019-02-26
Tutorial save design bug

sruddy:


packages/components/src/TokenTutorial/TokenTutorialLanding.tsx
```diff
@@ -93,7 +90,7 @@ export class TokenTutorialLanding extends React.Component<TokenTutorialLandingPr
                 <TutorialProgressText questions={topic.questions.length - lastSlideIdx} />
                 <TutorialLandingProgressBars>
                   {topic.questions.map((question, questionIdx) => {
-                    if (lastSlideIdx > 0 && questionIdx <= lastSlideIdx) {
+                    if (lastSlideIdx > 0 && questionIdx <= lastSlideIdx - 1) {
```
jorgelo: "There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors."

packages/components/src/Tokens/TokensAccountBuy.tsx
```diff
@@ -33,92 +33,77 @@ export interface TokenAccountBuyProps {
...
+export const UserTokenAccountBuy: React.StatelessComponent<TokenAccountBuyProps> = props => {
+  const { foundationAddress, network, faqUrl, step, onBuyComplete } = props;
+  let tokenSection;
 
-export class UserTokenAccountBuy extends React.Component<TokenAccountBuyProps, TokenAccountBuyStates> {
-  constructor(props: any) {
-    super(props);
-    this.state = { step: this.props.step };
-  }
+  if (step === TOKEN_PROGRESS.DISABLED) {
```
jorgelo: I think this is fine for now, but we should break these large render functions down into smaller ones, it'll be easier to read.

packages/components/src/Tokens/TokensAccountVerify.tsx
```diff
@@ -15,53 +17,70 @@ import { colors } from "../styleConstants";
...
+        </FlexColumnsPrimaryModule>
+      );
+    }
+
+    return <></>;
   }
 
-  return <></>;
-};
+  private isQuizStarted = (quizPayload: {}) => {
```
jorgelo: This could just be `return Object.keys(quizPayload).length > 0`.

(or even `return Object.keys(quizPayload).length`, but I like the explicit `> 0`).

packages/components/src/Tokens/TokensStyledComponents.tsx
```diff
@@ -101,25 +101,23 @@ export const FlexColumnsSecondaryModule = styled.div`
...
-  width: 100%;
-`;
 
-export const TokenHeader = styled.h2`
-  color: ${colors.primary.BLACK};
-  font-size: 27px;
-  line-height: 39px;
-  margin-bottom: 25px;
-`;
+  h2 {
```
jorgelo: I'm starting to really like using sub selectors in styled components 👍

packages/components/src/Tokens/TokensAccountBuy.tsx
sruddy: Agreed, this will need a refactor when I add the sell flow. I will break down the components in that ticket.

packages/components/src/Tokens/TokensAccountVerify.tsx
sruddy: 👍 good call

packages/components/src/Tokens/TokensStyledComponents.tsx
sruddy: Yes! I think it makes it a little cleaner and easier to read

PR #993, opened 2019-02-27
Airswap sell component

sruddy:
- Sell component
- temporary sell tab


packages/components/src/Airswap/Airswap.stories.tsx
```diff
@@ -27,4 +28,11 @@ storiesOf("Airswap", module)
         <AirswapBuyCVL network={"4"} buyFromAddress={"0x0..."} buyCVLBtnText={"Buy from the Civil Foundation"} />
       </Container>
     );
+  })
+  .add("Sell CVL", () => {
```
am9u: FYI, @tobek discovered that multiple `add()` calls in a single story file can cause the storyshot file sizes to balloon, so it might be worth it to create separate story files for each add call. 

(But this is not a blocker on this PR).

PR #1000, opened 2019-02-28
Login errors and font size changes

jorgelo:
* Fixes the font sizes for the auth flow.
* Adds error messages when the email is already in use, or when the email doesn't exist.


packages/components/src/styleConstants.ts
```diff
@@ -3,6 +3,7 @@ export const colors = {
     BLACK: "#000000",
     CIVIL_GRAY_1: "#5A5653",
     CIVIL_GRAY_2: "#7D7373",
+    CIVIL_GRAY_3: "#3F3C39", // Same as accent CIVIL_GRAY_0
```
sruddy: Let's keep this consistent and make it `CIVIL_GRAY_0` these go in order from darker to lighter

packages/components/src/Account/Auth/AuthStyledComponents.tsx
```diff
@@ -162,3 +165,27 @@ export const AuthEmailVerify = ({
...
+  border-radius: 4px;
+  background-color: #fff7f8;
+
+  background-position: 10px center;
+  background-image: url(${iconError});
+  background-size: 30px;
+  background-repeat: no-repeat;
+
+  color: #555555;
+  font-family: "Libre Franklin";
```
sruddy: Can you use the font style constant `font-family: ${fonts.SANS_SERIF};`

packages/components/src/styleConstants.ts
jorgelo: Good idea! Done.

packages/components/src/Account/Auth/AuthStyledComponents.tsx
jorgelo: Ooff, good catch.

packages/components/package.json
```diff
@@ -26,8 +26,9 @@
     ]
   },
   "scripts": {
-    "build": "yarn clean && tsc && mkdir build/images && cp -a src/images/. build/images",
-    "build:watch": "tsc -w",
+    "copy:images": "mkdir -p build/images && cp -a src/images/. build/images",
```
tobek: :pray: good call I always forget about this

packages/components/src/Account/Auth/AuthTextComponents.tsx
```diff
@@ -80,3 +80,22 @@ export const AuthTextSigninWithEmail: React.SFC = () => (
     </PageHeadingTextCentered>
   </>
 );
+
+// TODO(jorgelo): Toby, does hardcording the path here break anything on the newsroom?
```
jorgelo: Hey @tobek , I put a little note in here for you. Hardcoding these may break the newsroom signup flow if you're using the same components.

tobek: Thanks, yep I was just looking into this, it will break newsroom signup which uses `/apply-to-registry/signup|login`, but now I'm checking to see if it would be possible/easier to just use the `/auth/` flow with the redirect param, which in retrospect would have been the way to go in the first place but may a little more work to do now. Will take a look later today and either update to pass in the URL or move over to use one login flow.

One issue is that the newsroom signup flow has slightly different copy during email auth, so we'd probably have to pass an additional `context` param and vary copy based on that, or something. Will see.

PR #1002, opened 2019-02-28
Storefront sell screens

sruddy:


packages/components/src/Tokens/TokensTabSell.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const { isSellComplete } = this.state;
+
+    // TODO(sarah): temporary messaging while waiting on market maker
+    const comingSoon = true;
+    if (comingSoon) {
+      return (
+        <ComingSoon>
+          <h3>Coming Soon...</h3>
+          <p>
+            We appreciate your patience while we are testing this feature.<br />You will be notified when it’s ready. If
```
dankins: should we remove the "You will be notified when it’s ready." sentence?

sruddy: 👍 good call

PR #1010, opened 2019-03-01
Fix approve voting rights API wrapper method

am9u:


am9u:
Hmm, so with that suggested change, it looks like if the user tries to transfer a token amount that's less than the amount of tokens already in the voting contract, then no tokens get approved and the `requestVotingRights` transaction subsequently fails.

Here's some pseudo-code based on what's currently in `master` that I think demonstrates this.

```
// user already has 175 tokens in PLCR
approveForVotingRights(25); 
```

```
function approveForVotingRights(numTokens) {

const numTokensBN = ensureWeb3BigNumber(numTokens);  // 25
const currentApprovedTokens = await voting.getNumVotingRights();	 175
const difference = numTokensBN.sub(currentApprovedTokens); // -150
if (difference.greaterThan(0)) { // false, so nothing gets approved
}

}
```

It looks like the `approveForVotingRights` function was originally written to handle the Commit Vote flow, where the user specifies the total amount to commit to the vote aka the end number of tokens in the PLCR. 

This is a little different than the Transfer Tokens UI, where the user specifies an additional amount to add to the current PLCR balance. In order to re-use the same function, I made the update to add the current voting balance to the numTokens the user wants to transfer, which is similar to how the Commit Flow works.

Would it be better to create a separate approve function that will only be used by the Transfer Tokens UI that doesn't check the difference between voting balance and the numTokens arg?

PR #1011, opened 2019-03-01
Configurable Registry Alerts List ID

petersng:
Adds a configurable registry alerts list ID to be set by the environment in kubernetes.


packages/utils/src/email.ts
```diff
@@ -2,7 +2,8 @@ import * as querystring from "querystring";
 
 // TODO(jorgelo): Drop this into a configuration.
 const SIGNUP_ENDPOINT = "https://us-central1-civil-media.cloudfunctions.net/addToSendgrid";
-export const TCR_SENDGRID_LIST_ID = "5353193";
+export const TCR_SENDGRID_STAGING_LIST_ID = "5353193";
```
nickreynolds: so the PROD one is never used and the STAGING one is just for a story. Why don't we just delete these and use a magic string in the story?

PR #1016, opened 2019-03-01
Adds to mailing list

jorgelo:
* Adds a new user to the email list on sign up.


packages/components/src/Account/Auth/EmailAuth.tsx
```diff
@@ -13,12 +13,22 @@ import {
   ConfirmButtonContainer,
   AuthErrorMessage,
 } from "./AuthStyledComponents";
-import { isValidEmail } from "@joincivil/utils";
+import { isValidEmail, addToMailingList } from "@joincivil/utils";
```
petersng: Seems like `addToMailingList` is not needed here?

jorgelo: Argh, that's what I get for using auto add to import on VSCode. It automatically does this. I'll kill.

PR #1017, opened 2019-03-01
Works with query param

jorgelo:
* Instead of taking in the token as a url param `/verify-token/:token`, we take is as a query param `verify-token?jwt=xxx`


jorgelo:
@petersng When this is good to go, feel free to merge in right when the email template update goes out. (I don't want to break login / signup for everyone).

jorgelo:
@tobek Just realized that we'll need to do the same for the newsroom signup.

petersng:
@tobek is there an update on the corresponding work for newsroom signup?

tobek:
Dangit haha we used to use query string param and I changed it to match storefront URLs. I can revert my change, will do shortly, and then we can merge it and this PR and #125 all at once.

tobek:
Alright just created #1023 which will handle for newsroom signup.

PR #1021, opened 2019-03-04
update and store grant request data CIVIL-254

walfly:


packages/utils/src/apolloClient.ts
```diff
@@ -110,8 +109,18 @@ export function getApolloClient(httpLinkOptions: HttpLink.Options = {}): ApolloC
     }
   });
 
+  const createOmitTypenameLink = new ApolloLink((operation, forward) => {
```
tobek: :pray:

PR #1022, opened 2019-03-04
save last seen step to local storage CIVIL-399

walfly:


walfly:
It would be better to save this on graphql or something i think. but this will work for now

PR #1031, opened 2019-03-05
don't pop ethereum info when newsroom url clicked

nickreynolds:


nickreynolds:
Jon's PR took care of this

PR #1036, opened 2019-03-05
opt in tracking DRAFT

nickreynolds:
- use cookie to track
- add middleware dynamically when cookie found/set


packages/dapp/src/components/Main.tsx
```diff
@@ -39,22 +39,38 @@ import WrongNetwork from "./WrongNetwork";
...
+import { addMiddleware } from "redux-dynamic-middlewares";
+import { gaMiddleware } from "../redux/analytics";
 
 export interface MainReduxProps {
   network: string;
 }
 
 export interface MainState {
   prevAccount: EthAddress;
+  cookieSet: boolean;
```
am9u: Is this actually being used?

packages/dapp/src/components/Main.tsx
```diff
@@ -118,10 +134,25 @@ class Main extends React.Component<MainReduxProps & DispatchProp<any> & RouteCom
...
+  public allowTracking = (ev: any): void => {
+    this.answerTracking();
+    cookie.save("allowTracking", true, { path: "/" });
+    addMiddleware(gaMiddleware);
+    this.setState({ cookieSet: true });
+  };
+
+  public denyTracking = (ev: any): void => {
+    this.answerTracking();
+    cookie.save("allowTracking", false, { path: "/" });
```
am9u: I think we might want the `allowTracking` cookie to use the same `maxAge` value as the `didAnswerTracking` cookie. And maybe set that as a variable or constant in this file.

packages/dapp/src/components/Main.tsx
nickreynolds: kind of. it just modifies the state so we do a re-render and get rid of the AskTrackingBar

nickreynolds: but the value of it isn't used

nickreynolds: i guess `.forceUpdate()` is better, right?

packages/dapp/src/components/Main.tsx
nickreynolds: makes sense

packages/dapp/src/redux/store.ts
```diff
@@ -2,14 +2,31 @@ import thunk from "redux-thunk";
...
+export const store = () => {
+  const didAnswerTracking = cookie.load("didAnswerTracking");
+  if (didAnswerTracking) {
+    const allowTracking = cookie.load("allowTracking");
+    if (allowTracking) {
+      console.log("THIS ONE");
+      return createStore(
+        createRootReducer(history),
+        {},
+        compose(applyMiddleware(routerMiddleware(history), thunk, gaMiddleware)),
```
walfly: if this middleware isnt there what does trackpageview do?

nickreynolds: nothing, when the middleware isn't there no actions are tracked. when the middleware is added it picks them up again.

PR #1038, opened 2019-03-05
Buy eth notice

jorgelo:
* Adds a notice that informs the user they need to buy ETH to buy CVL
* Creates 3 versions of a Notice component
* Adds storybooks.


packages/components/src/Notice/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  border-radius: 4px;
+  background-color: #fff;
+`;
+
+export const NoticeContainer = styled<NoticeContainerProps, "div">("div")`
+  display: flex;
+  flex-direction: row;
+  background-position: 10px center;
+  background-size: 30px;
+  background-repeat: no-repeat;
```
tobek: Maybe I'm missing something but what's the purpose of background position/size/repeat when there's no background image?

jorgelo: @tobek Good catch, fixed.

PR #1039, opened 2019-03-05
Transfer token tab styling

sruddy:
Note this is just styling and doesn't include:
- unlock account message should be triggered off of unlockedWhitelist
- if the address is not on the civilianWhitelist they should be directed to complete the tutorial


packages/dapp/src/components/Dashboard/DashboardActivity.tsx
```diff
@@ -327,10 +338,24 @@ class DashboardActivity extends React.Component<
...
-              <DepositTokens />
+
+              {this.state.tutorialComplete ? (
+                <DashboardTransferTokenForm
+                  renderTransferBalance={this.renderTransferBalance}
+                  cvlAvailableBalance={balance}
+                  cvlVotingBalance={votingBalance}
+                >
+                  {this.state.transferTokenToVoting ? (
+                    <ReclaimTokens onMobileTransactionClick={this.showNoMobileTransactionsModal} />
```
am9u: These are switched. Reclaiming tokens is transferring FROM the PLCR and depositing is transferring TO the voting contract.

packages/dapp/src/components/Dashboard/DashboardActivity.tsx
```diff
@@ -339,6 +364,14 @@ class DashboardActivity extends React.Component<
     );
   };
 
+  private renderTransferBalance = (value: number) => {
+    if (value === 0 && this.state.transferTokenToVoting === false) {
```
am9u: Maybe we should use the `BalanceType` enum values in a state property (maybe call it `fromBalanceType` or `sourceBalanceType`?) here instead of this `transferTokenToVoting` boolean.  

Then this whole function could just be:

```
private renderTransferBalance = (value: number) => {
  this.setState({ fromBalanceType: value });
};
```

And the above ternary could be:

```
{this.state.fromBalanceType === BalanceType. AVAILABLE_BALANCE ? <DepositTokens /> : <ReclaimTokens />}
```

PR #1044, opened 2019-03-05
Copy changes

sruddy:


packages/components/src/TokenTutorial/TokenTutorialTextComponents.tsx
```diff
@@ -112,27 +109,13 @@ export const Topic1Tutorial4: React.SFC = props => (
...
-      Buying Civil tokens signals that you wish to participate in and contribute to Civil’s community. Civil tokens
-      unlock specific activities on the Civil platform, including sponsoring or launching a newsroom, challenging and
-      voting for/against Newsrooms for ethics violations or appealing the outcome of a community vote to the Civil
-      Council.
-    </p>
-    <p>
-      There will only ever be a fixed supply of Civil tokens created. Owners of Civil tokens will be economically
-      incentivized to ensure existing and future Civil Newsrooms maintain high quality standards as the network, and its
-      reach, grows.
+      For those that want to get more technical, CVL is a utility token based on the ERC20 protocol. It's a value stored
```
dankins: I assume this copy has all been vetted, but we say here CVL is a utility token and not a consumer token - which I think is a departure from what we normally say

sruddy: I flagged it in the quip doc and will update if needed

PR #1057, opened 2019-03-06
Donate Feature

jorgelo:
* Adds a button that sends a person to paypal to donate to the foundation.


packages/components/src/Tokens/TokensTextComponents.tsx
```diff
@@ -214,6 +214,14 @@ export const TokenETHFAQQuestion6Text: React.SFC = props => (
...
 );
 
+export const TokenDonateToCivilFoundationText: React.SFC = () => (
+  <>
+    <h3>Donate to the Civil Foundation</h3>
+    <p>Want to support this project, but don’t want to buy Civil tokens? </p>
+  </>
+);
+
+export const TokenETHFAQQuestion8Text: React.SFC = props => <h3>How do I unlock my tokens?</h3>;
```
jorgelo: @sruddy I may have messed up merging with master. Could you confirm this is what we meant to do?

sruddy: Yeah it looks like this is wrong. I think you can just remove `TokenETHFAQQuestion8Text` and you should move your text to the bottom I have everything grouped by component this looks like it's in the middle of the FAQ stuff

packages/components/src/icons/logos/Paypal.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <svg
+      width="64px"
+      height="16px"
+      viewBox="0 0 64 16"
+      version="1.1"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+    >
+      <g id="Symbols" stroke="none" stroke-width="1" fill="none" fillRule="evenodd">
+        <g id="Donation" transform="translate(-43.000000, -132.000000)">
```
sruddy: You should strip out ids from svgs and also make sure you fix `stroke-width`

packages/components/src/Tokens/TokensTextComponents.tsx
jorgelo: I removed `TokenETHFAQQuestion8Text`. Thanks!

packages/components/src/icons/logos/Paypal.tsx
jorgelo: Good call, cleaned up!

packages/components/src/icons/logos/Paypal.tsx
```diff
@@ -10,32 +10,13 @@ export const PaypalLogoIcon = (): JSX.Element => {
...
                   id="Shape"
```
sruddy: looks like there are ids in each of the `path` tags too

PR #1079, opened 2019-03-07
Tutorial fixes and improvements

sruddy:


packages/components/src/TokenTutorial/TokenTutorialLanding.tsx
```diff
@@ -81,14 +83,19 @@ export class TokenTutorialLanding extends React.Component<TokenTutorialLandingPr
...
+            activeTopic++;
+          }
 
-          // TODO(jorgelo): What do we do when isComplete is true (this means that this topic has been completed)
+          const topicDisabled = activeTopic === idx ? false : true;
 
           return (
             <TutorialTopic key={idx}>
-              <LaunchTopic onClick={() => this.openTutorial(idx, lastSlideNumber)} disabled={isComplete}>
+              <LaunchTopic onClick={() => this.openTutorial(idx, nextQuestionNumber)} disabled={topicDisabled}>
```
am9u: Just for my understanding, what is the issue with using `disabled={isComplete}`?

This new `topicDisabled` logic seems to say:

- if the current topic is complete, then the next topic is the active one. 
- if the current topic's index < active topic index (ie: wont this always be true if `isComplete === true`?), then topicDisabled = true

Am I missing something, or doesn't the above logic simplify to `topicDisabled === isComplete` or just using `isComplete` as the boolean?

sruddy: We only want the active topic to be enabled. So if you're on the second topic the first and third will be disabled. A topic doesn't exist in the payload till it is started so I can't just check for isComplete.

am9u: Ok, that make total sense when put into context! 

Maybe worth dropping that context into an inline comment so its clear for the future (ie: topic doesn't exist in payload until started so chan't check for `isComplete`) so this is well-communicated

PR #1080, opened 2019-03-08
Refactor and DRY up routing and links for the DApp

am9u:
This PR consolidates all the external links, as well as internal route definitions, for the dapp into its `constants.ts` file. This should let us more easily manage link changes going forward. 

Additionally, this update introduces the use of `formatRoute` for reverse generation of `react-router` routes based on route type and the route's expected properties.

Finally, this update consolidates various string variables used for URL generation into enum constants, with some more proper typing, with the intention of improving code maintenance.


am9u:
Added a wider range of reviewers, to keep you all updated with some changes to DApp routing.

Issue #979, opened 2019-02-25
Add Footer


nguyetv:
Design mock: https://consensys.invisionapp.com/share/CJQKH7KGQP8#/347067855_Navigation

For Developers
Github [link to https://github.com/joincivil]

Company
About us [https://civil.co/our-team/]
Blog [https://blog.joincivil.com/]
Civil Foundation [https://civilfound.org/]

Help
FAQ [https://cvlconsensys.zendesk.com/hc/en-us]
Contact [https://civil.co/contact/]

Legal
Terms [https://civil.co/terms/]
Privacy [https://civil.co/privacy/]

@CivilTCRAlerts [https://twitter.com/CivilTCRalerts]

PR #1084, opened 2019-03-11
Add the Footer to the DApp

am9u:
The PR adds a `Footer` component, with styles from the marketing site, and implements it in the DApp.

Issue: #979 

Final links still need to be confirmed by @nguyetv (ie: should the DApp have the same links as the marketing site? Or should they be unique for the dapp per the mock in this issue's ticket?)


packages/components/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,194 @@
...
+  currentYear: string;
+}
+
+export const Footer: React.SFC<FooterProps> = props => {
+  return (
+    <StyledFooter>
+      <StyledFooterInner>
+        <StyledFooterInnerLeft>
+          <StyledFooterLogo>
+            <a href="https://civil.co">
```
nickreynolds: don't we have a folder where icons live?

packages/components/src/Footer/Footer.tsx
```diff
@@ -0,0 +1,194 @@
...
+                </li>
+              </ul>
+            </div>
+            <div>
+              <ul>
+                <li>
+                  <b>About</b>
+                </li>
+                <li>
+                  <a href="https://blog.joincivil.com/" target="_blank">
```
nickreynolds: should all these links be constants / in an enum?

packages/components/src/Footer/Footer.tsx
am9u: Ah yeah, I just updated this PR with the proper icons (and added new ones where applicable).

packages/components/src/Footer/Footer.tsx
am9u: Yeah, that's a good point. I just got confirmation that these links are the correct ones, so I'll drop them into a constants file.

PR #1098, opened 2019-03-14
Apply to TCR + Transfer to Multisig step for Newsroom Signup

am9u:
This update implements the Apply to Registry step for Newsroom Signup's design and functionality, as well as the Application Submitted view, which includes the Add To Calendar functionality.

https://civilmedia.atlassian.net/browse/CIVIL-407

This PR also includes support for pulling the appropriate TCR parameter values for display in the UI's copy, as well populating web3 transactions with the proper arguments. Additionally, it copies the Dapp's higher order component for Transaction Status Modals in order to implement the correct Transaction Status Modal styles and behavior.


am9u:
Just for this draft/local testing. Will def be removing that change for the
final PR

packages/newsroom-signup/src/ApplyToTCR/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  userBalance: BigNumber;
+  multisigBalance: BigNumber;
+  applyStageLenDisplay?: string;
+  minDeposit: BigNumber;
+  multisigAddress?: EthAddress;
+  multisigHasMinDeposit: boolean;
+  listing?: ListingWrapper;
+}
+
+export type TApplyToTCRStepProps = ApplyToTCRStepOwnProps & ApplyToTCRStepReduxProps;
```
tobek: Just curious, is the `T` a typo or shorthand for "type"?

packages/newsroom-signup/src/ApplyToTCR/ApplyToTCRSuccess.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <OBSectionDescription>
+          You can also follow the progress of your application on your Newsroom Profile on the Registry.
+        </OBSectionDescription>
+
+        <Button size={buttonSizes.SMALL} href={registryListingURL}>
+          View your Newsroom
+        </Button>
+
+        <p>
+          If you have any questions, you can contact <a href="#">support</a>
```
tobek: Not sure if we have this in a constant anywhere but I think this can be support@civil.co.

packages/newsroom-signup/src/ApplyToTCR/index.tsx
am9u: `T` is a shorthand for "type". I think I picked this up from a TS blog somewhere.

am9u:
@tobek @walfly With the exception of some minor styling (padding, margins, some copy sizing), I think this PR is feature/design complete.

It's probably apparent from the code, but one thing to note for integration is the `address` prop for the `ApplyToTCR` parent component (the Newsroom's EthAddress). It's the key prop for the various UI states, so it needs to be defined when the component is rendered. But from my understanding, we should definitely have the newsroom's address by the time they're applying, so hopefully this note is trivial.

tobek:
This is all great, thanks for the hand off notes. I'm going to fix some merge conflicts and merge it in!

PR #1107, opened 2019-03-19
CIVIL-403 launch contract big refactor

walfly:


packages/newsroom-signup/src/actionCreators.ts
```diff
@@ -8,6 +8,7 @@ import { makeUserObject } from "./utils";
 
 export enum newsroomActions {
   ADD_NEWSROOM = "ADD_NEWSROOM",
+  ADD_MULTISIG = "ADD_MULTISIG",
```
tobek: Is this redundant with the `SET_MULTISIG_ADDRESS` action that jon's PR added?

walfly: yeah

PR #1113, opened 2019-03-20
add stand in page to skip adding newsroom to roster, change some cont…

walfly:
…rol flow stuff


packages/newsroom-signup/src/actionCreators.ts
```diff
@@ -68,11 +67,12 @@ export const getNewsroom = (address: EthAddress, civil: Civil, charter: Partial<
 ): Promise<AnyAction> => {
   const newsroom = await civil.newsroomAtUntrusted(address);
   const wrapper = await newsroom.getNewsroomWrapper();
+  console.log("here");
```
tobek: :raising_hand_man:

PR #1114, opened 2019-03-20
uniswap buy+sell

dankins:
This PR introduces uniswap buy and sell functionality, but puts it behind a feature flag. It can be enabled with by adding `?feature-flag=uniswap` to the URL. This uses a feature flag service, which I think we could eventually make more sophisticated to enable by user or whatever.

This has some text components that need to be pulled out, but I'll do that after we implement the updated designs.

This introduces `CivilContext`, which makes it easier to access services without having to pass it down the component heirarchy. There are many places where `getCivil()` is called - this can now be retrieved from the context (and components will update if that object changes).

Moving some components into `Tokens` subfolder (buy and sell) so they are a bit more organized.


packages/components/src/Tokens/sell/TokensTabSell.tsx
```diff
@@ -0,0 +1,59 @@
...
+    super(props);
+    this.state = {
+      isSellComplete: false,
+    };
+  }
+
+  public render(): JSX.Element | null {
+    const { isSellComplete } = this.state;
+    // TODO(sarah): check token controller.
+    const isTokenUnlocked = true;
```
sruddy: this should check the whitelist there's a helper in dapp helpers which should probably be moved to utils `getUnlockedWhitelist(userAccount)`

PR #1122, opened 2019-03-25
Newsroom signup copy updates [CIVIL-496] + style updates and using link constants

tobek:
Also @am9u @jorgelo @sruddy @nickreynolds just want to alert you to my updating utils `urlConstants` so that the FAQ URLs don't need to be prepended with `FAQ_BASE_URL` any more, they're all absolute links: https://github.com/joincivil/Civil/commit/71ae42af216e803faa41d63a4016ffc95a47212d#diff-c8363b3c529b70fdfc039948a2d161c8L546. If you make new FAQ links the old way they will not work.

### DIFF
--- a/packages/newsroom-signup/src/SmartContract/AboutSmartContractsButton.tsx
+++ b/packages/newsroom-signup/src/SmartContract/AboutSmartContractsButton.tsx
@@ -29,8 +29,8 @@ export class AboutSmartContractsButton extends React.Component {
               newsroom's identity.
             </Paragraph>
             <Paragraph>
-              Your index will appear in the Civil network, and provides a way to track it back to your website, along
-              with proof that the contents have not changed since last publish.
+              Using Civil's publishing tools, your content can be indexed on the Civil network, which provides a way to
+              track it back to your website, along with proof that the contents have not changed since last publish.
             </Paragraph>
             <Paragraph>
               Archiving will save the full text of your post to IPFS and, if you choose, the Ethereum blockchain. You
--- a/packages/newsroom-signup/src/SmartContract/AboutSmartContractsButton.tsx
+++ b/packages/newsroom-signup/src/SmartContract/AboutSmartContractsButton.tsx
@@ -1,0 +1,77 @@
+import * as React from "react";
+import { InvertedButton, buttonSizes } from "@joincivil/components";
+import { urlConstants } from "@joincivil/utils";
+import { InfoModalButton, Header, SectionHeader, Paragraph } from "../InfoModalButton";
+
+export class AboutSmartContractsButton extends React.Component {
+  public render(): JSX.Element | null {
+    return (
+      <InfoModalButton
+        buttonText="Learn more about Newsroom Smart Contracts"
+        content={
+          <>
+            <Header>About Newsroom Smart Contracts</Header>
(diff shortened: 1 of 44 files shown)


packages/newsroom-signup/src/SmartContract/UnderstandingEth.tsx
```diff
@@ -106,22 +103,24 @@ export class UnderstandingEth extends React.Component {
...
         <OBCollapsable
           open={false}
           header={<OBCollapsableHeader>How long does it take to buy ETH?</OBCollapsableHeader>}
         >
           <OBSmallParagraph>
-            If you purchase Ether (ETH) in Coinbase using your bank account, Coinbase uses a ACH (Automated Clearing
-            House, which is governed by the Federal Reserve) bank transfer system for U.S. customers, and it can take
-            3-5 business days for funds to transfer from your bank.
+            If you purchase Ether (ETH) in Coinbase using your bank account, Coinbase uses ACH. Automated Clearing
+            House, which is governed by the Federal Reserve is a bank transfer system for U.S. customers, and it can
```
walfly: grammatical issues here

packages/newsroom-signup/src/SmartContract/AboutSmartContractsButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+              see its activities on the blockchain. You cannot send funds directly to a newsroom smart contract.
+            </Paragraph>
+            <Paragraph>
+              In the case of Civil Registry and Publisher, this newsroom smart contract code allows you to create and
+              maintain a decentralized index and/or archive of your content and control who has permission to update it.
+              In combination with your newsroom wallet address, the smart contract acts as a verification for your
+              newsroom's identity.
+            </Paragraph>
+            <Paragraph>
+              Your index will appear in the Civil network, and provides a way to track it back to your website, along
```
walfly: does this really belong here,  also seems like an abrupt transition to bring up index and archive without introducing the terms

packages/newsroom-signup/src/SmartContract/UnderstandingEth.tsx
tobek: Good catch, missing comma, will fix.

packages/newsroom-signup/src/SmartContract/AboutSmartContractsButton.tsx
tobek: The previous paragraph gives context with "this newsroom smart contract code allows you to create and maintain a decentralized index and/or archive of your content" which makes sense (and is still pretty technically accurate) using the dictionary definitions of those terms. I would maybe update this sentence to say "Using Civil's publishing tools, your content can be indexed in the Civil network, which provides..." and then we're probably good without getting into the precise meanings of these terms in our context. @picocassette what do you think?

tobek:
For now I just made the copy updates that I suggested, can revisit if necessary.

packages/newsroom-signup/src/SmartContract/AboutSmartContractsButton.tsx
picocassette: Yeah, sounds good to me @tobek

PR #1127, opened 2019-03-26
add infura app key to config and use it for websocket provider

nickreynolds:


dankins:
just to confirm - we need to update the orchestration to set this env var right?

nickreynolds:
> just to confirm - we need to update the orchestration to set this env var right? couldn't we just update the entire URL rather than set a dependency on the infura key in the app?

https://github.com/joincivil/orchestration/pull/54

we could update the entire URL if you want and avoid any env stuff. if that's what you think we should do I can make that change

dankins:
Oh I guess I assumed that the ETH node was set via environment var, but it does not appear to be the case. then yea - the way you did it is :+1:

PR #1129, opened 2019-03-27
Integrate with steps API [CIVIL-408]

tobek:


packages/newsroom-signup/src/DataWrapper.tsx
```diff
@@ -81,16 +71,25 @@ export class DataWrapper extends React.Component<DataWrapperProps> {
...
-                              grantApproved,
-                              newsroomDeployTx,
-                              newsroomAddress,
-                              profileWalletAddress: data.currentUser.ethAddress,
-                              persistedCharter,
-                              saveAddress,
-                              persistCharter: this.saveCharterFuncFromMutation(saveCharter),
-                            });
+                            return (
+                              <Mutation mutation={saveStepsMutation}>
```
tobek: This now has 2 nested <Query>'s and 3 nested <Mutations>'s, needs refactoring with apollo's compose and HOCs, will do this after launch.

walfly: 👍

packages/newsroom-signup/src/ApplyToTCR/index.tsx
```diff
@@ -12,7 +12,7 @@ export interface ApplyToTCRStepOwnProps {
   address?: EthAddress;
   newsroom: any;
   civil?: Civil;
-  navigate(go: 1 | -1): void;
+  navigate(go: 1 | -1, allDone?: boolean): void;
```
walfly: can't devine this with like `step === numsteps`?

walfly: or `step > numsteps`

tobek: Yes, you are absolutely right (can just check `step = STEP.APPLIED`, simplifies the hack a bit.

Issue #793, opened 2019-01-18
Posting to IPFS transaction status modal needs styling tweaks


am9u:
![image](https://user-images.githubusercontent.com/143426/51398438-95046d00-1b11-11e9-816d-d5eaaa5d52bc.png)

PR #1133, opened 2019-03-27
Fix styling for Post to IPFS modals for the DApp

am9u:
Addresses #793

### DIFF
--- a/packages/dapp/src/components/utility/TransactionStatusModalsHOC.tsx
+++ b/packages/dapp/src/components/utility/TransactionStatusModalsHOC.tsx
@@ -153,10 +153,12 @@ export const hasTransactionStatusModals = (transactionStatusModalConfig: Transac
         (multiStepTransactionLabels && multiStepTransactionLabels[this.state.transactionType!]) || "1 of 1";
       const stepLabel = `Step ${stepLabelText} - ${transactionLabel}`;
       return (
-        <MetaMaskModal ipfsPost={true} waiting={true}>
-          <ModalStepLabel>{stepLabel}</ModalStepLabel>
-          <ModalHeading>Posting to IPFS</ModalHeading>
-        </MetaMaskModal>
+        <Modal width={558} textAlign="center">
+          <ProgressModalContentInProgress>
+            <ModalStepLabel>{stepLabel}</ModalStepLabel>
+            <ModalHeading>Posting to IPFS</ModalHeading>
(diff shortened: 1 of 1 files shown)


tobek:
This will also fix https://civilmedia.atlassian.net/browse/CIVIL-536!

packages/dapp/src/components/utility/TransactionStatusModalsHOC.tsx
```diff
@@ -153,10 +153,12 @@ export const hasTransactionStatusModals = (transactionStatusModalConfig: Transac
         (multiStepTransactionLabels && multiStepTransactionLabels[this.state.transactionType!]) || "1 of 1";
       const stepLabel = `Step ${stepLabelText} - ${transactionLabel}`;
       return (
-        <MetaMaskModal ipfsPost={true} waiting={true}>
-          <ModalStepLabel>{stepLabel}</ModalStepLabel>
-          <ModalHeading>Posting to IPFS</ModalHeading>
-        </MetaMaskModal>
+        <Modal width={558} textAlign="center">
+          <ProgressModalContentInProgress>
```
nickreynolds: this modal would give inaccurate info about what's happening - we could create a new container but it's not correct to say there's a tx happening here

packages/dapp/src/components/utility/TransactionStatusModalsHOC.tsx
```diff
@@ -153,10 +155,17 @@ export const hasTransactionStatusModals = (transactionStatusModalConfig: Transac
...
       const stepLabel = `Step ${stepLabelText} - ${transactionLabel}`;
       return (
-        <MetaMaskModal ipfsPost={true} waiting={true}>
+        <Modal width={558} textAlign="center">
+          <LoadingIndicator height={100} width={150} />
           <ModalStepLabel>{stepLabel}</ModalStepLabel>
           <ModalHeading>Posting to IPFS</ModalHeading>
-        </MetaMaskModal>
+          <ModalContentInsetContainer>
+            <ModalContent>This can take several minutes. Please don't close the tab.</ModalContent>
```
nickreynolds: we can update this to say "a few moments" or something since posting to IPFS is usually only like 10 seconds, not a blocker tho

am9u: Sure, I’ll update this in a PR this afternoon.

PR #1140, opened 2019-03-28
CIVIL-459 hide sign button if signed

walfly:


packages/newsroom-signup/src/Newsroom.tsx
```diff
@@ -289,17 +289,14 @@ class NewsroomComponent extends React.Component<NewsroomProps & DispatchProp<any
             account: this.props.account,
           }}
         >
-          <div ref={(el: HTMLDivElement) => (this.container = el)}>
-            <StepProcessTopNavNoButtons
-              activeIndex={STEP_TO_SECTION[this.state.currentStep]}
-              onActiveTabChange={this.navigateToSection}
-              contentPrepend={this.renderRepublishCharter()}
```
tobek: This line reverts an update of mine! Need that `contentPrepend`.

PR #1148, opened 2019-03-28
Fix hydrating the listing data after a successful Apply transaction

am9u:
This update attempts the following flow to ensure that the Apply to Registry step has the proper data needed to display on the success page, immediately following a successful Apply transaction

1) When getting listing data in the `getListing` action creator, first check the data returned from `getListingWrapper()`
2) If the data is good, hydrate the store with said data
3) If the data is bad (ie 0x0 owner address) and a listener doesn't currently exist, set up a listener for the listing's application events
4) When the application event comes in, call `getListingWrapper()` again, check the data, and if it’s good hydrate the store.
5) Unsubscribe the application listener


am9u:
Note: This update relies on an internal variable that encapsulates the current application event listeners, instead of creating a Redux reducer for this data. This is because of the amount of boilerplate to add a reducer is a bit high due to the fact that our redux action creators/reducers/provider definitions and implementation life in multiple places. I think it may be worth refactoring the Redux store into a separate package that can be used as a dependency by the various "view" packages (dapp, newsroom-signup, etc).

But, If anyone thinks this data really NEEDS to be in a reducer, feel free to say so. I can refactor this update, if we really feel strongly about this.

PR #1153, opened 2019-03-29
Storefront GA

jorgelo:
* Adds tracking for Token Storefront
* Adds generic tracking action.
* Fires tracking events for: Tokens purchased, logins, signups, quiz begin/end,


am9u:
Hmmm, yeah @tobek. The tighter binding of redux can make re-using these a little tricky. I agree that this is pretty good for now, but here are some thoughts on how to maybe refactor this for the future.

To be re-usable, I think the way to do this going forward may be to:

- Build these components so they expect some props functions which get called when tracking should be fired
- use `mapDispatchToProps` to create the prop functions by wiring up `dispatch` + the appropriate actions, and using redux's `connect` HOC to pass the prop functions into the view component
- The action generators, as well as the `mapDispatchToProps` function, could probably live in a separate package (alongside all the redux code?)
- This future "redux" package could contain all the defined reducers, action creators, and shared selectors for all Civil React+Redux apps. It could also have a `configStore` method that lets the implementing package specify whichever reducers the package needs, so it doesn't necessarily pull all of them.

Anyway, this seems like a larger refactor and definitely outside the scope of this PR, but perhaps a worthy project for the near future.

jorgelo:
@am9u @tobek Totally agree. Until I started this I didn't realize we never use `connect` in any of the components. I didn't want to tightly couple anything to redux right now, mainly because this felt like it should be a larger engineering discussion. That being said, I'm not against it. It was a real pain passing up all the events to the parent components.

PR #1156, opened 2019-04-01
upgrade to react 16.8 

dankins:
* it looks like we are already running 16.8, but our types were very old
* replaces deprecated `React.SFC` and `React.StatelessComponent` with `React.FunctionComponent`. this is like 98% of the changes in this PR
* there was typescript craziness going on with `packages/dapp/src/apis/civilTCR.ts`. got it to compile but this should require some additional functional testing. I think this is the only actual code change rather than just type changes
* changes `const enum` to just `enum`
* react-scripts-ts is deprecated, replaces with create-react-app https://vincenttunru.com/migrate-create-react-app-typescript-to-create-react-app


packages/components/src/Account/Auth/AuthStyledComponents.tsx
```diff
@@ -50,13 +50,13 @@ export const ConfirmedEmailLetterIcon = styled(CheckEmailLetterIcon)`
   background-image: url(${confirmedEmailImage});
 `;
 
-export const CenterWrapper: React.SFC = ({ children }) => (
+export const CenterWrapper: React.StatelessComponent = ({ children }) => (
```
am9u: Should these be `React.FunctionComponent`?

dankins: weird good catch. I did a global search and replace, not sure how that one got missed. This should still work, but will say its deprecated in when building. will fix

packages/dapp/src/apis/civilTCR.ts
```diff
@@ -33,31 +33,31 @@ export async function approveForApply(multisigAddress?: EthAddress): Promise<Two
...
   const appealFee = await government.getAppealFee();
-  return approve(appealFee);
+  return approve(appealFee, undefined);
 }
 
 export async function approve(
   amount: number | BigNumber,
-  multisigAddress?: EthAddress,
-): Promise<TwoStepEthTransaction | void> {
+  multisigAddress: EthAddress | undefined,
```
tobek: Shouldn't `multisigAddress?: EthAddress` functionally be the same as `multisigAddress: EthAddress | undefined`? Or is this different in a newer typescript version? Just tested locally and the former is more permissive - you can pass `undefined` or `EthAddress` or nothing at all, but in the latter you must explicitly pass `EthAddress | undefined`.

Seems like the former is more convenient (you can pass `undefined` if you want but you can just leave it off if you don't) and the resulting JS should be the same.

dankins: I was having trouble getting this to compile and it was just acting crazy (for example (`void | TwoStepEthTransaction` would compile but `TwoStepEthTransaction | void` wouldn't).

I think there might have been some weirdness with the resolve typescript version, which I think i sorted out after I made this change. I just went back and reverted any changes to that file and it appears to be fine. 

So yea, disregard.

PR #1158, opened 2019-04-01
uniswap rollout

dankins:
you can test this now at test.civil.app/tokens


packages/components/src/Tokens/sell/TokensTabSellActive.tsx
```diff
@@ -1,21 +1,86 @@
...
+            Don't see all of your tokens?
+          </a>
+        </div>
+        <FullScreenModal open={this.state.showModal} dismissOnOutsideClick={true} handleClose={() => this.hideModal()}>
+          <TokenInfoContainer>
+            <ModalHeading>Don't see all of your tokens?</ModalHeading>
+            <ModalContent>Check the amount of CVL in your voting balance.</ModalContent>
+            <ModalContent>
+              If your voting tokens are not used for any challenges, or votes, transfer them to your Available Balance
+              in order to sell. You can do this within the Dashboard section.
```
sruddy: We should link the the transfer token tab `/dashboard/tasks/transfer-voting-tokens` not sure if there is a url const yet for this

dankins: good call - just added

PR #1164, opened 2019-04-02
Fix nav to block users from certain steps

tobek:


packages/newsroom-signup/src/Newsroom.tsx
```diff
@@ -368,6 +360,37 @@ class NewsroomComponent extends React.Component<NewsroomProps & DispatchProp<any
...
+      currentStep = STEP.PROFILE_GRANT;
+    }
+
+    currentStep = this.backtrackSteps(currentStep);
+
+    return currentStep;
+  }
+
+  /** Handle situation where user used nav to jump too far ahead in step process before we put in disable checks - they should be backtracked back to where they need to be. */
+  private backtrackSteps(step: STEP): STEP {
```
walfly: does this have to be recursive, can't we just say `if there is no address and your current step is past launch contract then go back to the contract launch section` ?

packages/newsroom-signup/src/Newsroom.tsx
```diff
@@ -381,17 +404,22 @@ class NewsroomComponent extends React.Component<NewsroomProps & DispatchProp<any
   private getDisabled(section: SECTION): () => boolean {
     // @TODO/tobek Setting everything to enabled for now for testing, but we should work these out.
     const functions = {
-      [SECTION.CONTRACT]: () => {
+      [SECTION.PROFILE]: () => {
         return false;
       },
+      [SECTION.CONTRACT]: () => {
```
walfly: could include a charter check here, has required fields etc.  I guess that would also require error messaging

tobek: Because of our disabled checks inside newsroom profile stuff, they can't get to the grant page unless they've completed their charter, so checking they've passed grant step is inclusive of all the stuff before it. We have a bunch of charter logic in there https://github.com/joincivil/Civil/blob/master/packages/newsroom-signup/src/NewsroomProfile/index.tsx#L32-L67 that we can just reuse this way (unless I'm missing something).

packages/newsroom-signup/src/Newsroom.tsx
tobek: That would only handle the case where they clicked too far ahead before making a contract, but people could be anywhere. They could have started on the newsroom profile and immediately clicked on Apply To Registry nav step without having filled in charter, or they could have created contract but tried to skip tutorial, etc. Doing it recursively we get to reuse the logic in `getDisabled` in a granular way.

PR #1168, opened 2019-04-03
CIVIL-494 vary notice based on balance

walfly:


packages/newsroom-signup/src/SmartContract/UnderstandingEth.tsx
```diff
@@ -48,11 +72,7 @@ export class UnderstandingEth extends React.Component {
           buy Civil tokens later in this process.
         </OBSectionDescription>
 
-        <EthNotice type={NoticeTypes.ERROR}>
-          You can continue to the next step to create your Newsroom Smart Contract, but without ETH your transaction
-          will not go through. Please note, if you are a first-time purchaser, it may take a few days to get ETH in your
-          wallet.
-        </EthNotice>
+        <EthNotice type={NoticeTypes.ERROR}>{warningCopy}</EthNotice>
```
tobek: Should we also change the notice type if they have enough ETH? I think there's an "alert" one that would make sense.

PR #1170, opened 2019-04-03
CIVIL-485 and CIVIL-525 buy alert copy and buy complete unlock notice

dankins:
* CIVIL-485 'Unlock tokens" message still display for whitelisted, returned users
* CIVIL-525 Revise alert copy for "Buy" screen to buy ETH
* also sneaking in a couple of linting fixes

* TokensTabBuyComplete.tsx uses react hooks enabled buy the 16.8 upgrade


dankins:
Yea I definitely struggled with where to put that. My only reservation with `utils` is that utils is somewhat framework agnostic (ie, no dependency on React).

sruddy:
@dankins good point. I think there should be a better place for it but maybe this is something we can discuss when we redefine what components and storybook should be

Issue #1201, opened 2019-04-12
Change default Confirm vote message when users aren't logged in.


nguyetv:
When users aren't logged in to Metamask or wallet, the default message for Confirm vote says 

"You did not participate in this vote" 

Instead please change the message to say 

```Please connect your wallet to confirm your vote```
CTA: Connect Wallet

Issue #1195, opened 2019-04-10
Set "Add to calendar" event to Private


nguyetv:
Add copy under 'Add to my calendar' button to say 

```Remember to set event to private" 

design mockup: https://consensys.invisionapp.com/d/main/#/console/14237549/357402896/preview

Issue #1187, opened 2019-04-08
Implement v1.1 of 'Review my vote' screen


nguyetv:
Updated design: https://consensys.invisionapp.com/d/main/#/console/14237549/357402896/preview

The updated UI includes the following:
- New succinct intro copy
- one column layout with 'Add to calendar' button higher on the page
- a checkbox to remind users to save their info before Metamask can be activated. 
- Added a small transactions fee copy below Metamask button. 

confirm screen with "Add to Calendar" feature: https://consensys.invisionapp.com/d/main/#/console/14237549/347133791/preview

Current UI:
![Screen Shot 2019-04-08 at 8 03 53 AM](https://user-images.githubusercontent.com/6794566/55722744-119e3d80-59d5-11e9-884b-ec7c9e012af7.png)

am9u:
@nguyetv FYI, we are currently using a open source package for the 'Add to My Calendar' button, so we don't have much ability to customize styling. I think typography and colors may be doable, but getting the icon in there might be tough.

If we really want an icon in there, we'd prob have to write our own component for Add To Calendar, which might extend this ticket for a few more days.

Also, can you confirm that the date format in the Invision mocks is the one we want? We talked last week about keeping the numeral format, as it was clearer for localization purposes, but if this has changed, please let me know.

nguyetv:
@am9u Got ya. Don't worry about the icon then - it's not as important as the copy

And yes, the date format in the mockup is old. Please keep the current date formate. 

Thanks for checking :)

PR #1202, opened 2019-04-15
Updates to UI and copy to improve the DApp's Voting experience

am9u:
This PR includes the following updates:

- Update Challenge / Appeal Challenge Reveal Vote cards on Listing Detail page to show Confirm UI, not You Didn't Vote UI, by default. This should help alleviate some user confusion when the page loads first loads, especially because of the lag between loading the Listing data via GraphQL and User Challenge Data via Web3 (#1201)

- Change 'Enable Ethereum' text to 'Connect Wallet' in the Top Nav /
User Dashboard to be more user-friendly/less tech-speak

- Increase the height of the title/description portion of the Registry Homepage
cards to be 190px, to accommodate newsrooms with long names that span 3 lines ([CIVIL-607](https://civilmedia.atlassian.net/browse/CIVIL-607))

- Update Review Vote view per the latest design, which includes a checkbox that asks the user if they have saved their salt / voting information (#1187)

- Add text suggesting setting the Reveal Vote calendar event's visibility to Private to the Review Vote view (#1195)

### DIFF
--- a/packages/components/src/Account/Auth/__snapshots__/EmailAuth.stories.storyshot
+++ b/packages/components/src/Account/Auth/__snapshots__/EmailAuth.stories.storyshot
@@ -27,10 +27,10 @@ exports[`Storyshots Common / Auth / Email Signup Flow AccountEmailAuth 1`] = `
         />
       </div>
       <ul
-        className="sc-iKiVwC idmrno"
+        className="sc-emjYpo bOkvvA"
       >
         <li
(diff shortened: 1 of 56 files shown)


am9u:
@sruddy You bring up a good point. My original thinking was that this would be a bit more flexible should we want to re-use these components for some future unknown reason, but it is probably a bit unrealistic and less maintainable (as you noted). I'll update this PR per your feedback.

PR #1212, opened 2019-04-18
call to factory comes through another contract, origin should be used…

walfly:
… instead of sender


walfly:
Linter complains about using tx.origin here, but seems right in this case

dankins:
this seems fine, but are we sure we always want the *first* initial owner to be the editor? Is it possible for this to case weird results if someone didn't add themselves in the UI first ? (or if they don't need to add themselves I guess that would solve it) Could we use `tx.origin` in this case? I know it is generally frowned upon but I think it makes sense in this case.

walfly:
Tx.origin is bad and forbidden by the linter.  All owners should also be
editors, there are a number of transactions that are literally half price
if you don't have to go through the multisig.

I debated looping through initial owners, but there is literally nowhere in
our code where we send more than one.  So it didn't seem worth the gas.

dankins:
`All owners should also be
editors` <-- that makes sense

PR #1213, opened 2019-04-18
Improvements to handling post-grant transition [CIVIL-582]

tobek:
We've had a lot of support tickets about people stuck on the "waiting on grant" step. Some of these were from foundation not actually using the link that sets the "grant approved" flag in DB, but we've gotten at least two users for whom the flag is set correctly but they're still stuck. To combat this and track it down:

- Refactor "waiting on grant" logic to be simpler and more centralized
- Act as if grant was approved if the user has enough CVL to apply (most likely because they received grant)
- Add some GA events to help debug
- Also, replaces solo "continue" button on Smart Contract step with the same back/next buttons as every other step, since people are getting lost navigating and this could help a little

### DIFF
--- a/packages/newsroom-signup/src/Newsroom.tsx
+++ b/packages/newsroom-signup/src/Newsroom.tsx
@@ -1,4 +1,6 @@
 import * as qs from "querystring";
+import { BigNumber } from "bignumber.js";
+import { Parameters } from "@joincivil/utils";
 import {
   ButtonTheme,
   colors,
@@ -130,6 +132,7 @@ export interface NewsroomReduxProps extends NewsroomExternalProps {
   userIsEditor?: boolean;
   userNotOnContract?: boolean;
   charterUri?: string;
+  waitingOnGrant?: boolean;
 }
 
 export interface NewsroomGqlProps {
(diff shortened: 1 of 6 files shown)


packages/newsroom-signup/src/Newsroom.tsx
```diff
@@ -506,11 +533,22 @@ class NewsroomComponent extends React.Component<NewsroomProps & DispatchProp<any
   };
 }
 
-const mapStateToProps = (state: StateWithNewsroom, ownProps: NewsroomExternalProps): NewsroomReduxProps => {
+const mapStateToProps = (state: StateWithNewsroom, ownProps: NewsroomGqlProps): NewsroomReduxProps => {
   const { newsroomAddress } = ownProps;
   const newsroom = state.newsrooms.get(newsroomAddress || "") || { wrapper: { data: {} } };
+  const { user, parameters } = (state as any).networkDependent;
```
walfly: I don't know what to do about this, but this seems very wrong.  Reaching up out of this package into the parent context this way.

walfly: can this be passed in through root component

tobek: This is a good point, though we already do this in various spots in the tutorial, token purchase, and apply to TCR steps (e.g. https://github.com/joincivil/Civil/blob/a3bce8d935c4d9bc2d107132c3f823075999942d/packages/newsroom-signup/src/ApplyToTCR/index.tsx#L93)

Passing them in through root component would work but would have to be passed down the tree to a dozen child components. What we probably should do is add these to the Civil context or a new context.

It's a small-to-medium refactor, I think we should punt on this until such time that newsroom-signup appears in any other context or at least next time we need to use these values - would like to get this PR out since there are a few support threads going that this should help with.

walfly: i like the idea of using a context.  Can you leave a comment with a reminder about that or make a ticket

tobek: Yep will do

PR #1236, opened 2019-05-14
cast user challenge data query requests as bignumbers

nickreynolds:


nickreynolds:
pretty sure the linting errors here aren't related to my work

am9u:
I think `yarn prettier` just needs to be run, but it's fine. I can run it on my branch before setting a PR to `master`.

PR #1245, opened 2019-05-28
Import from package instead of relative path through monorepo

tobek:


tobek:
Ah that makes sense. Yeah I'm only catching these because it breaks when imported (as a nested dependency) into civil-sdk. Finding lots of little bits like this.

tobek:
Oops accidentally closed this.

PR #1249, opened 2019-05-31
Fix issue where user w/ locked/not installed MetaMask sees incorrect UI based on user challenge data

am9u:
- Check if user's EthAddress is available before requesting user challenge data from GraphQL

- Fix nested `<p>` elements issue with DApp Error Message component

### DIFF
--- a/packages/dapp/src/components/listing/ChallengeDetail.tsx
+++ b/packages/dapp/src/components/listing/ChallengeDetail.tsx
@@ -167,6 +167,10 @@ class ChallengeDetail extends React.Component<ChallengeDetailProps> {
     };
 
     if (this.props.useGraphQL) {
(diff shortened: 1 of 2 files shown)


packages/dapp/src/components/listing/ChallengeDetail.tsx
```diff
@@ -228,6 +232,8 @@ class ChallengeDetail extends React.Component<ChallengeDetailProps> {
     const canShowAppealChallengeRewardsFrom =
       didUserCommit(userAppealChallengeData) && this.props.challenge.appeal!.appealChallenge!.resolved;
 
+    console.log(inCommitPhase, userChallengeData);
```
nickreynolds: console log left in

PR #1257, opened 2019-06-03
Boost and Boost Feed Pages

sruddy:


packages/dapp/src/components/Boosts/Boost.tsx
```diff
@@ -0,0 +1,44 @@
...
+  boostId: string;
+}
+
+class BoostPage extends React.Component<BoostPageProps> {
+  public render(): JSX.Element {
+    return (
+      <>
+        <Helmet title="Civil Boost - The Civil Registry" />
+        <ScrollToTopOnMount />
+        <FeatureFlag feature={"boost"} replacement={<ComingSoonText />}>
```
tobek: Do we want this coming soon text? Hadn't seen anything about it but I might have missed it. The default replacement is `<></>` which would also work.

Also, the feature flag nick mentioned in slack which I used in mine is `boosts-mvp`

sruddy: You can only reach this page if you know the url but I think it's better to say coming soon than have a completely blank page with no messaging

tobek: Oh yeah good point, that makes sense!

PR #1283, opened 2019-06-24
Boosts nav menu items + newsroom check [CIVIL-739]

tobek:


packages/components/src/NavBar/NavMenu.tsx
```diff
@@ -44,6 +48,23 @@ const NavMenuLinksComponent: React.FunctionComponent<NavMenuCloseDrawerProp> = p
           <NavLinkContractAddressesText />
         </NavLink>
       </NavDropDown>
+      <FeatureFlag feature={"boosts-mvp"}>
```
nickreynolds: not important right now, but next time we use a feature flag we should create a constant for it

PR #1284, opened 2019-06-25
Boost page styles

sruddy:


packages/dapp/src/components/Boosts/Boost.tsx
```diff
@@ -19,14 +19,9 @@ class BoostPage extends React.Component<BoostPageProps> {
...
-          <BoostHeaderWrapper>
-            <BoostHeader>
-              <h1>Civil Boost</h1>
-              <a href="/boosts">&lsaquo; Back to Boost</a>
-            </BoostHeader>
-          </BoostHeaderWrapper>
           <BoostWrapper>
-            <Boost open={true} boostId={this.props.boostId} />
+            {/* TODO(sruddy) check if user is boost owner */}
+            <Boost open={true} boostId={this.props.boostId} boostOwner={false} />
```
sruddy: @tobek `boostOwner` is a similar check to what you were talking about in Slack let me know if you are already working on it

tobek: Yep I'm working on this for edit boost. Depending on how it fleshes out I might be able to take over this TODO easily, or else when it's done I'll send it over to you to check out and use here.

PR #1288, opened 2019-06-27
Edit Boost page [CIVIL-736]

tobek:
Build will fail until https://github.com/joincivil/civil-sdk/pull/25 merged and version bumped here.


tobek:
This is weird. Why isn't this build failing? This PR relies on updates to civil-sdk component props - some added, some deleted, some renamed. I guess linting doesn't understand React props, but I'm surprised that dapp build didn't fail on this.

PR #1292, opened 2019-07-01
Publish

tobek:
- @joincivil/artifacts@1.1.2
 - @joincivil/components@1.9.1
 - @joincivil/contracts@4.8.2
 - @joincivil/core@4.8.2
 - @joincivil/dapp@0.9.2
 - @joincivil/dev-utils@1.6.2
 - @joincivil/ethapi@0.4.2
 - @joincivil/newsroom-manager@1.8.2
 - @joincivil/newsroom-signup@1.8.2
 - @joincivil/typescript-types@1.4.2
 - @joincivil/utils@1.9.2


tobek:
I wish! `lerna publish` is supposed to do exactly that but it doesn't seem to be working and I couldn't get to the bottom of it. Whenever I run `lerna updated` it shows all packages have been updated. I've tried the `from-git` and `from-package` publish flags but either I'm doing it wrong or they don't work either.

Some of this might be because of dependencies - if `components` or `core` is updated then every package depending on it gets updated even if the package itself hasn't changed, because lerna keeps version numbers at the latest across all packages in the workspace. However, `artifacts` doesn't have any dependencies but it keeps thinking it has been updated too.

PR #1294, opened 2019-07-02
Newsroom multisig withdraw [CIVIL-816]

tobek:


packages/core/src/contracts/newsroom.ts
```diff
@@ -319,6 +319,21 @@ export class Newsroom extends BaseWrapper<NewsroomContract> {
...
+   * @param to Address to transfer to
+   */
+  public async transferEthFromMultisig(
+    eth: BigNumber,
+    to: EthAddress,
+  ): Promise<TwoStepEthTransaction<MultisigTransaction>> {
+    const wei = toWei(eth, EthereumUnits.ether);
+    const address = await this.multisigProxy.getMultisigAddress();
+    const multisig = await Multisig.atUntrusted(this.ethApi, address!);
+    return multisig.submitTransaction(to, wei, "");
```
tobek: Not sure if sending `""` payload is normal but it works. Multisig `submitTransaction` function signature requires a payload, because it passes it to `sendTransactionAsync` in autogenerated multisig wrapper, which requires a payload, and which would be a hassle to change, so I'm guessing this is ok?

nickreynolds: should be fine. it's a tx with no data

PR #1305, opened 2019-07-10
Adds Boost to listings tab

sruddy:
I added a feature flag to the boost but I can't add it to the tab without breaking tabs. If we don't want Boost in the tab with a coming soon message we can wait on this.

### DIFF
--- a/packages/dapp/src/components/listing/ListingRedux.tsx
+++ b/packages/dapp/src/components/listing/ListingRedux.tsx
@@ -22,6 +22,7 @@ import EmailSignup from "./EmailSignup";
 import ListingOwnerActions from "./ListingOwnerActions";
 import ListingDiscourse from "./ListingDiscourse";
 import ListingHistory from "./ListingHistory";
+import ListingBoosts from "./ListingBoosts";
 import ListingHeader from "./ListingHeader";
 import ListingCharter from "./ListingCharter";
 import ListingPhaseActions from "./ListingPhaseActions";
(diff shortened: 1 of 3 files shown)


nickreynolds:
I think it's fine to give a little 4 day "preview" of what's coming...

packages/dapp/src/components/listing/ListingRedux.tsx
```diff
@@ -172,6 +173,12 @@ class ListingPageComponent extends React.Component<
                 </ListingTabContent>
               </Tab>
 
+              <Tab title="Boosts">
```
am9u: What happens if you wrap `<FeatureFlag>` around this tab?

sruddy: Yes, if I wrap it here the tab isn't clickable and it breaks the styles:

Screen Shot 2019-07-10 at 9 35 35 AM

tobek: I'd say let's wait until we enable the feature flag on prod for this.

I wonder if it's fixable though. Here's the line in FeatureFlag: https://github.com/joincivil/Civil/blob/master/packages/components/src/features/FeatureFlag.tsx#L17 - I'm not sure if the `children` prop can be directly output without wrapping it in fragment element, but we could give it a prop that takes a `JSX.Element` and just output that (I'm not sure why we're not doing that with `replacement`).

PR #1327, opened 2019-07-20
MetaMask login for the DApp

am9u:
This is a first and basic pass at adding Web3 login to the DApp. The next iteration will have some better styling that more closely matches our mockups.

- Add basic Sign Up / Login with MetaMask functionality to the Dapp
- Includes some updates to access Apollo Session info from localstorage as a React state variable (via a hook)
- Includes basic redirect parameter support (`?next=[redirect url]`), since we no longer need to persist a redirect url to the email magic link 
- Fix for nested `<a>` on the Registry homepage


nickreynolds:
will take a closer look at this once it's ready for review, but one thing is that I'm not sure we actually need separate flows for login vs signup. There must be a way where we can just submit the signed message and the backend creates the account if necessary and gives auth token, right? and then we just direct user to add email + username if they don't have those?

am9u:
@nickreynolds So the signup/login mutations are currently two very specific and separate functions as to which action they try to perform. They expect specific prefixes in their respective signed messages, as well as seem to do separate lookup/create logic. Since they currently are separate, I thought it would be best to build the first pass of web3 login using what we currently have.

That said, it would be great if we could consolidate those two mutations into a single `ethAuth` mutation. Would love to get the BE engineering thoughts on that. @dankins @petersng

dankins:
It is trivial to do on the backend, but I have concerns with this approach, since it can lead to creation of multiple accounts and cause a lot of confusion. For example, you might have your non-primary account selected and you try to log in, but it instead creates a new account.

nickreynolds:
maybe a compromise would be that if you try to login with a wallet that doesn't have an account, it won't create one (but would direct you to signup flow) but if you try to signup with a wallet that already has an account, it will tell you you already have one and just log you in?

packages/dapp/src/components/Auth/index.tsx
```diff
@@ -142,11 +179,11 @@ export class AuthRouter extends React.Component<RouteComponentProps> {
     });
   };
 
-  public handleOnAuthenticationContinue = (isNewUser: boolean): void => {
+  public handleOnAuthenticationContinue = (isNewUser: boolean, redirectUrl?: string): void => {
```
nickreynolds: I think we should set the account (and kick off action to subscribe to token balances) in redux here (and remove all the account stream subscriptions stuff in Main.tsx)

am9u: @nickreynolds I like the idea of co-locating the account/balances dispatchers with the Auth logic, but this method only gets called when the user actually logins via the login UI. So we would still need a place to kickoff those account/balance subscriptions for users who already have a valid session.

For the sake of this PR being reviewable and discrete, I'll look at refactoring this code, including moving the account/balance subscriptions here, in an upcoming PR that'll land this week.

PR #1329, opened 2019-07-22
exit listing

nickreynolds:
- update listing owner actions so exit listing also withdraws tokens to user wallet after exiting listing transaction
- updates needed to `ownerAddresses` in graphql newsroom query (otherwise need to disable graphql to see owner actions)


nickreynolds:
@tobek @picocassette - we should make sure to add this action to the new newsroom manager ui (and remove this tab from the listing, if it's taken care of somewhere else). Should also add other simple token actions (deposit/withdraw to/from multisig, deposit/withdraw to/from listing)

PR #1334, opened 2019-07-23
Connect Stripe account to newsroom channel [CIVIL-888] [CIVIL-882]

tobek:
Note @sruddy that I made a `withNewsroomChannel` HOC that we can use to get from newsroom address to channel ID in a couple lines: https://github.com/joincivil/Civil/pull/1334/files#diff-5a60fd730cae493532d3d239e10bae14R33


packages/components/src/TCRUserDashboard/DashboardNewsroomStripeConnect.tsx
```diff
@@ -0,0 +1,201 @@
...
+            proceeds.
+          </p>
+          {this.renderStripeConnectButton()}
+        </>
+      );
+    }
+  }
+
+  private renderStripeConnectButton(linkText?: string): JSX.Element {
+    // @TODO/toby Our stripe client ID (which should be different on production) should be in some config, but we don't have any config for components package, would have to pass in from dapp via prop or add to civil context.
```
tobek: @dankins or anyone else, how should we do this? We could define in https://github.com/joincivil/Civil/blob/master/packages/dapp/.env or we could pass in as env variable into react app the way we are doing for `SENTRY_DSN`.

dankins: this likely needs to be passed in the same way as SENTRY_DSN, I can make updates to the `orchestration` repo to set this. We could probably tack the entire `config` object into CivilContext so you can grab it from stuff in `components`

tobek: Updated using this approach (passing in config through context). Added the client id to orchestration in https://github.com/joincivil/orchestration/pull/95/ but put the staging id in dapp `.env` too so that it works locally.

packages/components/src/TCRUserDashboard/DashboardNewsroomStripeConnect.tsx
```diff
@@ -191,10 +194,12 @@ export class DashboardNewsroomStripeConnectComponent extends React.Component<
   }
 
   private renderStripeConnectButton(linkText?: string): JSX.Element {
-    // @TODO/toby Our stripe client ID (which should be different on production) should be in some config, but we don't have any config for components package, would have to pass in from dapp via prop or add to civil context.
-    const url =
-      "https://connect.stripe.com/oauth/authorize?response_type=code&client_id=ca_BzqgUsw7tnnCpVaoQ157mrxtuCdN7h2q&scope=read_write&redirect_uri=http%3A%2F%2Flocalhost%3A3000%2Fdashboard%2Fnewsrooms";
-    return <p>{linkText ? <a href={url}>{linkText}</a> : <StripeConnectButtonLink href={url} />}</p>;
+    // @TODO/toby Remove `stripe-admin` feature flag query string param when launched
```
dankins: alternatively you could just pop the feature flag into the .env file (`REACT_APP_FEATURE_FLAGS`) - they are overridden at build time for staging and prod anyways but allows it to work locally

tobek: Oh sweet yeah I saw those and was going to ask to make sure they get overridden on staging/prod.

Thing is though if we do want to, say, test on prod with the feature flag not enabled by default yet, we still need to pass feature flag in to redirect url we send to stripe in order to test it so that this component is loaded when you get back.

PR #1342, opened 2019-08-05
Newsroom Manager update charter UI [CIVIL-994]

tobek:
- Reuses `newsroom-signup` components in new "manage newsroom" page, reworking them with copy/design updates to support this new context
- Misc `newsroom-signup` updates and improvements
- Designs are in flux, this is a barebones version. It's clean and nothing looks broken, but it doesn't match the designs.
- UI is behind `newsroom-channels` flag for now.

### DIFF
--- a/packages/newsroom-signup/src/NewsroomManager/index.tsx
+++ b/packages/newsroom-signup/src/NewsroomManager/index.tsx
@@ -3,7 +3,17 @@ import { connect, DispatchProp } from "react-redux";
 import styled, { ThemeProvider } from "styled-components";
 
 import { EthAddress, CharterData, NewsroomInstance } from "@joincivil/core";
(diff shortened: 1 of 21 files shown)


packages/newsroom-signup/src/NewsroomManager/index.tsx
```diff
@@ -0,0 +1,254 @@
...
+  LoadingMessage,
+  LoadingIndicator,
+  metaMaskLoginImgUrl,
+} from "@joincivil/components";
+
+import { StateWithNewsroom } from "../reducers";
+import {
+  getNewsroom,
+  updateCharter,
+  // analyticsEvent,
```
walfly: should get rid of the commented out stuff

tobek: Will do

PR #1360, opened 2019-08-15
Newsroom channel HOC support `newsroomContractAddress` prop name too

tobek:
So that we can use this HOC on `BoostForm` in civil-sdk which uses `newsroomContractAddress` prop name.


dankins:

PR #1363, opened 2019-08-16
Support newsroom channels from Launch Boost nav dropdown [CIVIL-1022]

tobek:
Without this, Launch Boost from nav is broken, so I think we need to deploy this to prod before we have anyone else use boosts (otherwise we'd have to make sure to direct them to launch boost via dashboard).

### DIFF
--- a/packages/dapp/src/components/Dashboard/ManageNewsroom/WithNewsroomChannelAdminList.tsx
+++ b/packages/dapp/src/components/Dashboard/ManageNewsroom/WithNewsroomChannelAdminList.tsx
@@ -6,7 +6,7 @@ import { EthAddress } from "@joincivil/core";
 import { LoadingMessage } from "@joincivil/components";
 import gql from "graphql-tag";
 
-const newsroomChannelAdminQuery = gql`
+export const newsroomChannelAdminQuery = gql`
   query {
     currentUser {
       uid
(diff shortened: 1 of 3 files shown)


packages/dapp/src/components/Dashboard/ManageNewsroom/WithNewsroomChannelAdminList.tsx
```diff
@@ -33,6 +33,16 @@ export interface WithNewsroomChannelAdminListProps {
   children(props: { newsroomAddresses: Set<EthAddress> }): any;
 }
 
+export function newsroomChannelsFromQueryData(data?: any): EthAddress[] {
+  if (data && data.currentUser && data.currentUser.channels && data.currentUser.channels.filter) {
+    return data.currentUser.channels
+      .filter((channelMember: any) => channelMember.channel.channelType === "newsroom")
```
nickreynolds: don't we also need to filter on whether or not the channelMember is an admin?

tobek: Good question! I thought that the `currentUser { channels { ... } }` query only returned channels that the current user is an admin of, but if not then yes we should filter by `currentUserIsAdmin`. Can you or @dankins confirm the behavior?

nickreynolds: Just checked the query and I don't see anything that only returns channelMember results of type admin. But since your query already gets the `role` you can just check if that is `admin`, I believe

dankins:

PR #1371, opened 2019-08-21
fix divide by zero bug

dankins:


dankins:
I agree, but I am just re-exporting the ethers.utils function and thought it would be more confusing if I renamed it. This way all of the downstream packages use the same ethers version as specified in `@joincivil/typescript-types`

dankins:

PR #1374, opened 2019-08-21
commitlint and semantic release

dankins:
this doesn't actually release anything yet, but sets up the framework so we can add what we need one at a time.


dankins:

PR #1375, opened 2019-08-21
release sdk

dankins:
this doesn't actually move civil-sdk into the monorepo yet, but just does an npm release. I think a lot of the packages are going to complain once they need a release that the tag already exists - the fix would be to just put a commit that bumps the version. I will run through all of them once I get a sdk set up properly

this also moves test/lint into the "build" job. I noticed that a bunch of time in CI is step in between job steps, and persisting workspaces. Combining them cut ~3 minutes off of the build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,26 +24,12 @@ jobs:
             - run:
                 name: Building
                 command:  yarn build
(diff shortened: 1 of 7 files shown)


dankins:

PR #1391, opened 2019-08-29
fix(newsroom-signup): greatThanOrEqualTo should be gte

dankins:
this was the old bignumber.js api and should be using bn.js
this didn't get caught since it the type was "any".


dankins:
:tada: This PR is included in version 4.8.14 :tada:

PR #1400, opened 2019-09-16
Standalone Boost embed in dapp [CIVIL-1047]

tobek:
To test this, run `yarn start` in dapp and go to http://localhost:3000/boost-embed?boost=5a7005c9-f46d-45fb-bdd8-c782be7489dc

***

I tried several methods for Boost embeds. Instead of creating a new bundle and new entry point, this PR adds a slot inside many of our more general dapp wrappers (config, civil context, apollo provider) but outside wrappers that are unnecessary for Boost embeds (no user/web3 stuff, no `Main.tsx`, no header/footer).

The exact location of embeds in the wrapper hierarchy may need tweaking. Do we want to be inside `<LoadUser>`? Do we want to load `<Web3AuthWrapper>`? Do we need to initialize any registry contracts? But it won't be hard to move it around and/or extract some stuff from `Main.tsx` and lift it higher.

The reason I did not do a separate bundle and entry point entirely:

1. It was really hard. It would either require ejecting create-react-app or doing something very complicated with `react-app-rewired` that I couldn't quite get to work (my efforts at https://github.com/joincivil/Civil/compare/tobek/boost-embed-entry)
2. The embed does need a lot of dapp wrapper stuff in `index.tsx` and `App.tsx` - using a separate entry point we'd have to abstract all that stuff into helpers or repeat code.
3. Separate entry point didn't save us much bundle size. With the approach in this PR we still save a bit from not loading any async route chunks, and we still get faster page load due to not having to execute contract code and other stuff. Gzipped total resources loaded on production build:
    - dapp homepage: 2.2mb
    - dapp boost permalink page: 2.1mb
    - boost embed as new entry/bundle: 1.7mb
    - boost embed inside dapp: 2.0mb **<-- this PR**

The only downside is that if we properly split up our various packages, we'll probably find there are some things we want to remove from embed but not from dapp. But a) there is lower-hanging fruit that benefits both, and b) I believe we could still achieve a separation here using async component chunks.


packages/dapp/src/App.tsx
```diff
@@ -42,18 +44,25 @@ export class App extends React.Component {
...
-                    </>
-                  );
-                }}
-              </LoadUser>
+              <Switch>
+                {standaloneRoutes.map(route => (
+                  <Route key={route.pathname} path={route.pathname} component={route.component} />
+                ))}
+                <Route>
+                  <LoadUser>
```
nickreynolds: I think the boost entry point should be inside LoadUser, we're gonna want to have user session data eventually

dankins: FWIW `civilUser` is moving into civil context so this component won't be needed after the Kirby PR

tobek: Is that this PR? https://github.com/joincivil/Civil/pull/1401 if so I'll leave it where it is for now.

dankins: yep
https://github.com/joincivil/Civil/pull/1401/files#diff-027941cdab4a902d3a0dcb788bebf089R18

dankins:
:tada: This PR is included in version 4.8.17 :tada:

PR #1401, opened 2019-09-17
feat(components): context refactor

dankins:
Sorry for the massive PR. This branch started as Kirby integration, but turned into a bit of beast since I needed to refactor anywhere that was calling "getCivil()" (which was essentially a global variable). This PR adds the `CivilProvider` component, which puts `civil` and some other goodies into context. It then adds `CivilHelper` context underneath that in the dapp. `src/apis/civilTCR.ts` was a boat load of exported functions that called `getCivil`, so that has been refactored into a class that accepts a Civil instance and puts the object into context. 

Also fixing a bunch of linting issues that were complaining about legacy context:
`public static contextType: React.Context<ICivilContext> = CivilContext;`
becomes:
```
  public static contextType = CivilContext;
  public context!: ICivilContext;
```

There were a couple places using `CivilContext.Consumer`, so I changed them to use hooks (`React.useContext`).

ethapi gets a little update since it was previously polling for account and network changes, this just adds listeners to the provider. This works for Metamask and Kirby.

This adds Kirby behind a feature flag, but doesn't really work after the username refactor. I figured I would at least get this code in and then adjust with Kirby so that everyone can start using the new code and prevent crazy merge issues later.


packages/components/src/TransactionButton.tsx
```diff
@@ -157,9 +158,8 @@ export class TransactionButtonNoModal extends React.Component<TransactionButtonP
 
   private onClick = async () => {
     const { civil } = this.context;
-
     if (civil && civil.currentProvider) {
-      await civil.currentProviderEnable();
+      // await civil.currentProviderEnable();
```
nickreynolds: add a comment explaining why this is commented out?

dankins: this was a mistake - added it as part of testing kirby and missed cleaning it up

dankins:
:tada: This PR is included in version 4.8.17 :tada:

PR #1419, opened 2019-10-01
feat(component): pew components

sruddy:


packages/components/src/Leaderboard/Leaderboard.tsx
```diff
@@ -0,0 +1,32 @@
...
+  LeaderboardItem,
+  LeaderboardAvatar,
+  LeaderboardUserName,
+  LeaderboardAmount,
+} from "./LeaderboardStyledComponents";
+import { Contributers } from "./types";
+
+export interface LeaderboardProps {
+  label?: string;
+  contributers: Contributers[];
```
nickreynolds: maybe I'm misunderstanding what this component is (tried to run storybook to look at it, but couldn't), but I would've thought that the Leaderboard would take in 3 separate Contributors (specified as like `contributorFirstPlace`, etc) rather than just an array of contributors. If this is the full list (and not just the 1st, 2nd, 3rd) can we rename the prop to like `sortedContributors`?

sruddy: @nickreynolds I'm not sure if there is already a data schema for this but I think `contributorFirstPlace` etc is kind of limiting. The leaderboard design is used in various ways maybe it's just the name `Leaderboard` that's misleading. Sometimes it's the top 3, sometimes the 3 most recent or maybe we would want the most frequent contributors someday. We may also want to do 5 or 15 someday. I think it make more sense to be able to search on the list and just pass the data you want to a more generic component instead of having several very specific components. Also easier if we ever want to A/B test variations.

sruddy: I will change the prop name though

nickreynolds: yeah I think the name is what threw me off, change that and we're good

dankins:
:tada: This PR is included in version 4.8.22 :tada:

PR #1422, opened 2019-10-01
feat(dapp): story feed

sruddy:


packages/components/src/StoryFeed/Story.tsx
```diff
@@ -13,11 +12,11 @@ export const Story: React.FunctionComponent<StoryProps> = props => {
     <>
       <StoryLink onClick={props.handleOpenStory}>
         <StoryLinkLeft>
-          <StoryTitle>{props.title}</StoryTitle>
-          <TimeStamp>{props.timeStamp}</TimeStamp>
+          <StoryTitle>{props.openGraphData.title}</StoryTitle>
+          <TimeStamp>{props.openGraphData.article.published_time}</TimeStamp>
```
nickreynolds: I don't think we should use the open graph published time here, we should use the post created-at time

packages/components/src/StoryFeed/StoryFeed.stories.tsx
```diff
@@ -16,6 +16,31 @@ const contributors = [
...
+  },
+};
+
+const openGraphData = {
+  article: {
+    published_time: "2019-08-28T15:58:32Z",
+  },
+  description: "Uyghurs are gaming TikTok’s algorithm to find a loophole in Xinjiangs’s information lockdown",
+  images: {
+    url: "https://codastory.com/wp-content/uploads/2019/09/Still-Header-.png",
```
nickreynolds: if this were to match how it's stored on the backend, images is actually an array of objects. so accessing the image would be images[0].url

packages/components/src/StoryFeed/types.ts
```diff
@@ -0,0 +1,28 @@
...
+    };
+  };
+}
+
+export interface OpenGraphData {
+  article: {
+    published_time: string;
+  };
+  description: string;
+  images: {
```
nickreynolds: left this comment elsewhere, but makes more sense here. if this were to match how it's stored on the backend, images is actually an array of objects. so accessing the image would be images[0].url

packages/components/src/StoryFeed/Story.tsx
sruddy: @nickreynolds Is post created at the time of link submission or the time the article was published?

sruddy: there's a conversation in slack pod-pulse about which we should use

dankins:
:tada: This PR is included in version 4.8.23 :tada:

PR #1423, opened 2019-10-02
feat(components): kirby

dankins:
this one is pretty big, sorry. I am creating this as a draft because I still need to add some more documentation and small cleanup tasks, but this is ready for review for the most part

* moves registry into its own "app"
* adds kirby "app"
* elements 
* adds elements package
* components starts to use elements
* removes semantic.min.css 
* removes dev-utils subprovider


dankins:
> * not entirely sure what the purpose of the "Stories" app is
my thinking here was we will use this as the entry point to the Storyfeed. this ensures that none of the Registry files so we can minimize time to First Meaningful Paint

dankins:
:tada: This PR is included in version 4.8.23 :tada:

PR #1424, opened 2019-10-03
feat(components): payments component

sruddy:


packages/components/src/Payments/PaymentsEth.tsx
```diff
@@ -0,0 +1,177 @@
...
+  PayWithEthDescriptionText,
+} from "./PaymentsTextComponents";
+import { PaymentsModal } from "./PaymentsModal";
+import { PaymentsEthForm } from "./PaymentsEthForm";
+
+export enum MODEL_CONTENT {
+  WHY_ETH = "why eth",
+  WHAT_IS_ETH = "what is eth",
+  CAN_USE_CVL = "can I use cvl",
+}
```
tobek: If you don't care about the values you can use this sugar and just give the key names like here https://github.com/joincivil/Civil/blob/master/packages/components/src/Notice/index.tsx#L6-L12 (I think it just uses the key name as the value?)

packages/components/src/Payments/PaymentsEth.tsx
```diff
@@ -0,0 +1,177 @@
...
+  userAddress?: EthAddress;
+  usdToSpend: number;
+  isWalletConnected: boolean;
+  handlePaymentSuccess(): void;
+}
+
+export interface PaymentsEthStates {
+  isMobileWalletModalOpen: boolean;
+  isInfoModalOpen: boolean;
+  modalContent: string;
```
tobek: Minor, but this type could be `MODEL_CONTENT` (I guess `modalContent?: MODEL_CONTENT`) so that you can set it to `undefined` for no modal showing.

packages/components/src/Payments/PaymentsEthForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+          amount: this.props.etherToSpend,
+          usdAmount: this.props.usdToSpend.toString(),
+          emailAddress: this.state.email,
+        },
+      },
+    });
+  };
+
+  private onTransactionError = (err: string) => {
+    this.context.fireAnalyticsEvent("tips", "ETH support rejected", this.props.postId, this.props.usdToSpend);
```
tobek: Super minor, but this analytics text makes it seem like maybe the user rejected the metamask modal, which also would result in an error (unless that's being handled in `TransactionButton` modals elsewhere? if it's not it would be nice to abstract that some day) but other things could cause an error too, so I'd say "error" instead of "rejected".

packages/components/src/Payments/PaymentsInputValidationUI.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    ${mediaQueries.MOBILE} {
+      width: 100%;
+    }
+  }
+
+  select {
+    -moz-appearance: none;
+    -webkit-appearance: none;
+    appearance: none;
```
tobek: Just checked and according to https://www.styled-components.com/docs/basics styled-components handles vendor prefixes for us :pray:

packages/components/src/Payments/PaymentsEthForm.tsx
sruddy: I brought this over from the Boost tracking Nick set up to keep it consistent. It would be good to check that out though and see how clear it is.

dankins:
:tada: This PR is included in version 4.8.24 :tada:

PR #1429, opened 2019-10-04
fix(dapp): add boost embed back in refactored dapp architecture

tobek:
A bit hacky because boost embed should go inside `RegistrySection` (it reuses no less than 6 of the registry wrapper components) BUT if we are on the boost embed then we don't want to show the fallback `RegistryShell` nav bar (and whatever else that contains in the future) while it's loading.

It's a small hack, but open to any other ideas about how to do this or how to make more explicit.


dankins:

PR #1430, opened 2019-10-04
feat(dapp): support analytics event in boost embed [CIVIL-1035]

tobek:
Refactors a bit to create yet another dapp wrapper component. Moved a bunch of the logic from `RegistrySection` out into `RegistryWrapper`, in order to cleanly use dispatch and context in `RegistrySection` to instantiate the fire analytics event early enough that the boost embed can share that logic.

Look at https://github.com/joincivil/Civil/pull/1429 first.

### DIFF
--- a/packages/dapp/src/registry/RegistryWrapper.tsx
+++ b/packages/dapp/src/registry/RegistryWrapper.tsx
@@ -1,0 +1,89 @@
+import * as React from "react";
+import { ApolloProvider } from "react-apollo";
+import { getApolloClient } from "@joincivil/utils";
+import config from "../helpers/config";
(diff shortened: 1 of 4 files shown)


packages/dapp/src/registry/RegistryWrapper.tsx
```diff
@@ -0,0 +1,89 @@
+import * as React from "react";
```
tobek: Nothing changed here, whole file is made up of stuff cut from `RegistrySection.tsx`.

tobek:
Yeah I was totally thinking the same, but yeah we don't know exactly which wrappers a future component might want to reuse so I figured we can leave for now and abstract out when we need to.

PR #1428, opened 2019-10-04
fix(dapp): clean up login/signup style

dankins:
also logs the user out if they change their metamask account

PR #1432, opened 2019-10-04
load nav async

dankins:
(draft until https://github.com/joincivil/Civil/pull/1428 is merged)
this PR might be easier to review by commit rather than as a whole since it has a few distinct changes.

* moves a bunch of icons into `elements` from `components`
* upgrade react-router
* couple kirby bugfixes
* the big change is moving the header and footer out of the components package and into the dapp. it loads the nav and footer, then lazy loads the "UserAccount" section. this results in a much snappier experience. 
* this also removes the "connect wallet" button - it looks to see if the user account is set and renders that portion. this makes your avatar appear much quicker rather than waiting for kirby / web3 and all that jazz to propegate


dankins:

PR #1435, opened 2019-10-07
Boost embed style updates [CIVIL-1079] [CIVIL-1031]

tobek:
Adds `renderContext` attribute to `CivilContext` which can be (currently) set to `DAPP` (default) or `EMBED` so that we can conditionally render some Boost stuff based on that context. Also adds that value in `styled-components` theme

Style updates:

- Disable input field when boost is ended
- Prevent scrolling in embed + show fading-out "view boost" link
- Add border to embed
- Link boost newsroom name to newsroom
- Better responsiveness for boost share widget
- Hide "visit newsroom" link on embed
- Truncate boost title length in embed

### DIFF
--- a/packages/elements/src/containers/mediaQueries.ts
+++ b/packages/elements/src/containers/mediaQueries.ts
@@ -1,5 +1,6 @@
 export const mediaQueries = {
-  MOBILE: "@media only screen and (min-width: 320px) and (max-width: 1030px)",
+  MOBILE: "@media only screen and (max-width: 1030px)",
+  MOBILE_SMALL: "@media only screen and (max-width: 640px)",
   COARSE_POINTER: "@media (pointer: coarse)",
   HOVER: "@media (hover: hover)",
 };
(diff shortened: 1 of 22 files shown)


packages/elements/src/containers/mediaQueries.ts
```diff
@@ -1,5 +1,6 @@
 export const mediaQueries = {
-  MOBILE: "@media only screen and (min-width: 320px) and (max-width: 1030px)",
+  MOBILE: "@media only screen and (max-width: 1030px)",
```
tobek: Removed the min 320px here, I think there are very very few mobile devices that are narrower than 320px, but if there were, it should still get mobile styles, right? Not sure why this was in there.

tobek:
@dankins let me know if you have any thoughts on this commit https://github.com/joincivil/Civil/pull/1435/commits/b3f64b169bb4b902e2fb6766e8b4959d8c6adc5e for adding `renderContext` to `CivilContext`, relevant to the refactors you've been doing.

dankins:

PR #1438, opened 2019-10-09
fix(dapp): set up analytics in AppProvider so boost embed can use again

tobek:
Moves analytics event setup into AppProvider, via new AnalyticsProvider, so that embeds can use. This was done in a clunky way in #1430 but was removed by refactor in #1432, but the refactor supports a simpler way to implement it.


dankins:

PR #1439, opened 2019-10-10
fix(dapp): workaround to fix payment route in boost embed [CIVIL-1107]

tobek:
Dumping my notes/slack convo about this issue for posterity, but the TL;DR is just to read the code.

according to react-redux ConnectedRouter docs, if you use it you have to remove any use of BrowserRouter or else there will be these problems syncing route state. this wasn't an issue before but with upgrade from react-router v4 to v5 it's now an issue: boost embed component doesn't receive updates when route changes and no use of useParams or withRouter anywhere in the component tree works.

so one fix that works is to replace the top level BrowserRouter with ConnectedRouter, which only works when wrapped in redux Provider, so we have to lift that too, and all the redux store code, which shouldn't need much, but in practice at the moment it ends up bringing in the whole codebase and borks all the code splitting progress.

also considered getting rid of ConnectedRouter, which we only use for one thing right now, which is to automatically dispatch route change events to google analytics for pageviews, not sure how much a refactor it would be to change how that works

we could also perform the top-level App.tsx split between kirby/stories/embed/registry just using document.location instead of a router, so that we can load ConnectedRouter further down

we could also downgrade react router

for now i found a workaround where useRouteMatch works instead of useParams if you hard code the path you're on. don't know why, i assume it relies on router logic that isn't clobbered by the BrowserRouter/ConnectedRouter issue. not sure how widespread this issue will be, if it blossoms then downgrading react router will be the better call but will leave this for now

### DIFF
--- a/packages/dapp/src/constants.ts
+++ b/packages/dapp/src/constants.ts
@@ -1,3 +1,7 @@
+export enum embedRoutes {
+  BOOST = "/embed/boost/:boostId/:payment?",
+}
+
 export enum routes {
   REGISTRY_HOME = "/registry/:listingType/:subListingType?",
   REGISTRY_HOME_ROOT = "/registry",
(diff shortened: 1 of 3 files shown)


dankins:

PR #1440, opened 2019-10-10
fix(sdk): get web3 working in boost embed [CIVIL-1107]

tobek:
In the dapp registry app we call this from the global nav, wasn't getting called properly here. Also with this update, ethereum enable will only be called from embed once the user enters the payment flow.


tobek:
Will merge now cause this is needed for testing boost embed payment flow, but @dankins you can check it out.

The question here is where to call ethereum enable I guess? For registry calling from global nav is obvious, for embeds we could call from top level EmbedsApp but we prob don't want to pop up MM window as soon as embed loads, but rather when they enter payment flow.

There's also the question of how kirby provider enable works inside an embed, mostly style-wise.

dankins:

PR #1442, opened 2019-10-11
Tons more embed tweaks so more fits without scrolling + misc tweaks

tobek:


packages/components/src/CurrencyConverter/UsdEthConverter.tsx
```diff
@@ -29,7 +30,7 @@ export const UsdEthConverter = (props: UsdEthConverterProps) => {
               currencyLabelFrom="Enter USD Amount"
               currencyCodeTo="ETH"
               currencyLabelTo="Amount of ETH"
-              displayErrorMsg={true}
+              displayErrorMsg={props.displayErrorMsg !== false}
```
nickreynolds: `displayErrorMsg={props.displayErrorMsg}` doesn't work?

tobek: No, because then if you supply `undefined` (as every other instance of this component does) then error messages would be disabled. Only want to disable on an explicit `false`. Other option would be to pass it directly but set the default value for that prop to be `true` - a few more lines of code but would be more explicit. Going to merge this now to get it out but can v easily change.

dankins:

PR #1444, opened 2019-10-11
Fixes and improvements for republish charter flow

tobek:
Fixes the two issues affecting republish charter:

1. `newsroom.updateRevisionURIAndHash` passed default `signature` argument to web3 call as `""` which errored on "invalid bytes value" - `"0x"` instead fixed that.
2. Sometimes IPFS charter publish times out, but there's no feedback so UI just hangs. This catches IPFS (also any web3) errors and displays an error modal.

Also some other misc fixes.

### DIFF
--- a/packages/core/src/contracts/newsroom.ts
+++ b/packages/core/src/contracts/newsroom.ts
@@ -686,7 +686,7 @@ export class Newsroom extends BaseWrapper<NewsroomContract> {
     contentId: ContentId,
     uri: string,
     hash: string,
-    signature: string = "",
+    signature: string = "0x",
   ): Promise<TwoStepEthTransaction<RevisionId | MultisigTransaction>> {
     if (contentId === 0) {
       await this.requireOwner();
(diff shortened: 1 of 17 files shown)


packages/core/src/contracts/newsroom.ts
```diff
@@ -686,7 +686,7 @@ export class Newsroom extends BaseWrapper<NewsroomContract> {
     contentId: ContentId,
     uri: string,
     hash: string,
-    signature: string = "",
+    signature: string = "0x",
```
tobek: @nickreynolds and/or @dankins want to run this by you. So empty string here was causing this error in charter republish:

    invalid bytes value (arg="signature", coderType="bytes", value="", version=4.0.27)

(Same error was breaking newsroom withdraw functionality with same fix, I think it's from web3 update.)

Sanity check: does it make sense to replace every empty string in this file that gets passed to a web3 call with `"0x"` as I've done in this PR? This is just `author` and `signature` arguments. None of the other functions here are actually used in the dapp, so I can't test them without rebuilding and firing up the plugin again which would be an ordeal right now, but if we leave them alone then they're potential landmines.

We have a bunch of empty strings default arguments in civilTCR contract calls too, but I haven't followed the calls down to see if they're passed to web3 calls in the same way, but if we run into a similar issue in the future this is the first thing we should check.

dankins: maybe just create a jira ticket to remind us to follow up but otherwise just tackle the issues as they appear in the wild

dankins:

PR #1445, opened 2019-10-11
chore(sdk): don't call ethereum enable in boost until ETH pay flow selected

tobek:
Previously `BoostPayEth.tsx` called `civil.currentProviderEnable` on mounting so you would get kirby (or MM) popup as soon as you entered payment flow even if you hadn't yet selected ETH or USD. This way it only pops up if you continue down ETH path.

Also fixes bug with currency converter telling you you don't have enough ETH if ethereum is enabled after the component is mounted.


dankins:

PR #1447, opened 2019-10-11
feat(kirby): disable metamask login/signup if no injected web3 found

nickreynolds:
- remove incognito


packages/kirby/src/views/identity/Login.tsx
```diff
@@ -38,6 +38,9 @@ export const Login: React.FunctionComponent<RouteComponentProps> = () => {
     (ctx.core.plugins.view as ViewPlugin).completeView();
   }
 
+  const metaMaskSubheading = hasInjectedWeb3 ?
+    "Your Browser has Web3 built-in! Click here to use your built-in Web3 wallet." :
```
dankins: imo should be "browser".

PR #1453, opened 2019-10-14
fix(kirby): show web3enable view on saved preference

dankins:
if logged out of portis from another window and then reload the page the portis login screen will be displayed from the iframe, but the kirby view is not displayed so the user cannot do anything.

### DIFF
--- a/packages/kirby/src/views/Web3Enable.tsx
+++ b/packages/kirby/src/views/Web3Enable.tsx
@@ -1,39 +1,66 @@
 import * as React from "react";
-import {
-  LogInWithMetaMask,
-  LogInWithPortis,
-  LogInWithBurner,
-  CoreContext,
-  useSelector,
-  CenteredPage,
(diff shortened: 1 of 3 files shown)


packages/kirby/src/views/Web3Enable.tsx
```diff
@@ -8,32 +11,60 @@ import {
...
+      (ctx.core.plugins.view as ViewPlugin).completeView();
+    }
   }
+
+  if (status === "provided") {
+    return <></>;
+  } else if (status === "enabling provider") {
+    return (
+      <CenteredPage>
+        <ClipLoader />
```
nickreynolds: can we add some text here like how we have "waiting for wallet connection" and "waiting for signuatre"?

dankins: good call, added

dankins:

PR #1458, opened 2019-10-14
fix(sdk): workaround for currentProviderEnable promise issue

tobek:
Instead of awaiting on `civil.currentProviderEnable()`, just call it without awaiting and then await on `this.context.civil.accountStream.first().toPromise()` instead


tobek:
No longer needed cause https://github.com/joincivil/Civil/pull/1461

PR #1459, opened 2019-10-14
style(dapp): restore slim mobile header

tobek:
Mobile header used to be 54px high, was changed recently to be taller and so it overlapped the top of page content. Could move all the page content down but it looked fine making the header 54px again, with a couple tweaks.


dankins:

PR #1465, opened 2019-10-15
feat: story boost and payment updates

sruddy:


packages/components/src/Payments/PaymentsAmount.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+export interface PaymentsAmountProps {
+  newsroomName: string;
+  suggestedAmounts: SuggestedAmounts[];
+  handleAmount(usdToSpend: number, shouldPublicize: boolean): void;
+}
+
+export interface PaymentsAmountStates {
+  showInput: boolean;
+  shouldPublicizeChecked: boolean;
```
tobek: It looks like this state isn't being used.

packages/elements/src/icons/PhotoDragIcon.tsx
```diff
@@ -2,27 +2,12 @@ import * as React from "react";
...
-                <path
-                  d="M29.715415,12.5355731 C22.9268775,12.5355731 17.4031621,18.0592885 17.4031621,24.8478261 C17.4031621,31.6363636 22.9268775,37.1600791 29.715415,37.1600791 C36.5039526,37.1600791 42.027668,31.6264822 42.027668,24.8379447 C42.027668,18.0494071 36.5039526,12.5355731 29.715415,12.5355731 Z M29.715415,35.3715415 C23.9051383,35.3715415 19.1818182,30.6482213 19.1818182,24.8379447 C19.1818182,19.027668 23.9051383,14.3043478 29.715415,14.3043478 C35.5256917,14.3043478 40.2490119,19.0375494 40.2490119,24.8379447 C40.2490119,30.6482213 35.5256917,35.3715415 29.715415,35.3715415 Z"
-                  id="Shape"
-                ></path>
-                <path
-                  d="M15.2885375,15.2529644 L11.5335968,15.2529644 C11.0395257,15.2529644 10.6442688,15.6482213 10.6442688,16.1422925 C10.6442688,16.6363636 11.0395257,17.0316206 11.5335968,17.0316206 L15.2885375,17.0316206 C15.7826087,17.0316206 16.1778656,16.6363636 16.1778656,16.1422925 C16.1778656,15.6482213 15.7826087,15.2529644 15.2885375,15.2529644 Z"
-                  id="Path"
-                ></path>
-              </g>
-            </g>
```
tobek: Removing all these nested transforming `<g>`s makes sense but a heads up that I did see that once changing the positioning of the icon and I had to adjust spacing around it. We should remove them to begin with so we only have to arrange the positioning once, cc @nickreynolds, but I think in the case of avatar upload any small translation changes will be fine, it's mainly only tricky when the icons have to be inline with text or something.

sruddy: We should ask design to optimize before sending to us

dankins:

PR #1471, opened 2019-10-18
fix(sdk): call eth check even if boost eth pay is mounted selected [CIVIL-1142]

tobek:
The error was kind of dumb and unrelated to my investigations.

Previously this only checked wallet in componentDidUpdate transition from not selected (i.e. choosing between when eth and credit card options) to selected. If stripe isn't enabled for a boost, component mounts with eth option already selected so it wasn't getting called.

Also has tiny mobile responsive css fix.


dankins:
:tada: This PR is included in version 4.8.30 :tada:

PR #1475, opened 2019-10-21
Min $2 for embedded project boost card payments + small tweaks

tobek:
Because of how usdToSpend is stored in history state it was slightly harder (or maybe I'm missing something) to do what story boost flow does and just update payment amount to $2 if they continue, so instead I added a link to go back and edit payment amount, which isn't as nice, but it's kind of an edge case, and we'll replace this flow at some point anyway.


dankins:

PR #1478, opened 2019-10-23
feat: story share

sruddy:


packages/elements/src/share/ShareStory.tsx
```diff
@@ -0,0 +1,131 @@
+import * as React from "react";
+import styled from "styled-components";
+import { ShareEmailIcon, LinkIcon, ShareTwitterIcon } from "../icons";
+import { colors } from "../colors";
+import { fonts } from "../text";
+
+const ShareWrapper = styled.div`
+  width: 300px;
```
tobek: Maybe an edge case but this should probably have max width 100% - on smallest mobile viewport 320px, with page padding there might be less than 300px for this guy and it'd overflow.

sruddy: 👍 I meant to add a media query to this

dankins:

PR #1479, opened 2019-10-23
feat: mobile avatar selection updates

nickreynolds:
- proper resizing of avatar selection component
- misc updates in navbar & modal
- cleanup


packages/dapp/src/components/header/styledComponents.tsx
```diff
@@ -35,6 +35,7 @@ export const NavOuter = styled.div`
     min-height: 54px;
     padding-top: 0;
     padding-bottom: 0;
+    max-width: 100%;
```
tobek: I'm still seeing some issues with this - my not-that-long username `tobytest2` is cut off and the dropdown arrow to expand isn't visible.

image

We don't have to put too much work into this edge case and can't accommodate arbitrary length usernames but i think it would make sense to get rid of some of the extra space here and have a `mediaQueries.MOBILE_SMALL` query to crop username length.

Also think we should white-space: nowrap the buttons

image

I'm playing around with this on my end and have the CSS working in dev tools so I can make a followup PR with some tweaks.

nickreynolds: we only need to support 15 character usernames, I think it can be done without cutting it off. We could hide the avatar on super-small screens if necessary

dankins:

PR #1480, opened 2019-10-24
Reduce unnecessary/premature kirby popups on Boosts [CIVIL-1134] [CIVIL-1186]

tobek:
Primarily a refactor of `BoostPermissionsHOC` to use channel admin gql query to check whether user should have permission to edit a boost, instead of web3.


dankins:

PR #1491, opened 2019-10-25
fix: test stripe payment request fix

nickreynolds:
-  untested, but only way to easily test is to put it on staging. plus it's broken already so wouldn't be breaking anything anyways


dankins:

PR #1513, opened 2019-11-04
Story boost embed + embed script

tobek:
- Create `/embed/boost/story/<boost-id>` route for story boost embed
- Create script that queries for ExternalLink and if it doesn't exist, creates it, and then creates iframe pointing to dapp story boost embed
- Remove @joincivil dependencies from embed loader - only React and Apollo now
    - This necessitated duplicating a little bit of logic and code from utils and components
    - Potentially both React and Apollo could be removed. It would complicate the code somewhat and make stuff less reusable. With them, the embed script is ~100kb gzipped which isn't bad at all.
- Copy iframe + wrapper from project boost
- Handle loading and error states
- Update @joincivil/sdk scripts to better test embed
- Lots of other things

This can be run with `yarn sdk:start:boost-embed` from sdk package (it will serve actual story boost embed from staging unless you updated `dappOrigin` in `embed.tsx` to point to local dapp).


packages/dapp/src/embeds/StoryBoostLoader.tsx
```diff
@@ -0,0 +1,88 @@
...
+const StoryBoostLoaderComponent: React.FunctionComponent = () => {
+  const civilContext = React.useContext<ICivilContext>(CivilContext);
+  civilContext.renderContext = RENDER_CONTEXT.EMBED;
+  const theme = {
+    ...DEFAULT_CHECKBOX_THEME,
+    ...DEFAULT_BUTTON_THEME,
+    renderContext: RENDER_CONTEXT.EMBED,
+  };
+
+  // Due to a conflict between react-router v5's `BrowserRouter`, which we use, and react/redux `ConnectedRouter`, which we also use (and which would take an unknown/large refactor to change without breaking code splitting gains), neither `useParams` hook nor `withRouter` are receiving updates here, so we have to use `useRouteMatch` and manually provide the boost embed route. - @tobek
```
dankins: is there a Jira to track this? probably should put it on our radar to fix eventually

packages/dapp/src/embeds/StoryBoostLoader.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const { boostId } = useRouteMatch<StoryBoostLoaderParams>(embedRoutes.STORY_BOOST)!.params;
+
+  return (
+    <EmbedWrapper>
+      <CivilLogoLink href="https://registry.civil.co" target="_blank">
```
dankins: can this be put in a constant? this is going to change once we launch storyfeed (registry.civil.co -> civil.co)

packages/dapp/src/embeds/StoryBoostLoader.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <CivilIcon />
+      </CivilLogoLink>
+      <ThemeProvider theme={theme}>
+        <StoryBoost
+          boostId={boostId}
+          isLoggedIn={false}
+          handleLogin={() => {
+            alert("@TODO/tobek");
```
dankins: This is fine since StoryBoost doesn't have login functionality right now right?

packages/sdk/src/react/StoryBoost/embed.tsx
```diff
@@ -0,0 +1,152 @@
...
+      id
+    }
+  }
+`;
+
+// const dappOrigin = "http://localhost:3000"; // if simultaneously working on/testing dapp embed locally
+const dappOrigin = process.env.NODE_ENV === "development" ? "https://staging.civil.app" : "https://registry.civil.co";
+const fallbackFallbackUrl = dappOrigin + "/storyfeed"; // if embed doesn't load, suggest they go here
+
+const apolloClient = new ApolloClient({
```
dankins: I assume this one-off apolloClient is so that we didn't need to bring in a bunch of other packages? (utils maybe?)
We should think about how to refactor this at some point

packages/dapp/src/embeds/StoryBoostLoader.tsx
tobek: There is! https://civilmedia.atlassian.net/browse/CIVIL-1121

That comment is copy-pasted from previous `BoostLoader.tsx` - the two files could be DRY'd up a lot, but I'm not yet sure what all actually needs re-using as things are still a bit in flux as I navigate implementation issues.

packages/dapp/src/embeds/StoryBoostLoader.tsx
tobek: Yep good call will do.

packages/dapp/src/embeds/StoryBoostLoader.tsx
tobek: It does actually, there's just some logic that handles passing these in from `StoryFeed` component that I still need to abstract out so I can use it here.

packages/sdk/src/react/StoryBoost/embed.tsx
tobek: Yeah this is to avoid bringing in utils which brings in a ton of stuff. We could definitely break out apollo utils into its own package which would do the trick. Though I did just realize I could change this so that we don't need apollo here at all - all the queries could be moved to inside the iframe if instead of loading embed by boost ID, we just load e.g. `/embed/boost/story?url=...` and then all the gql action can happen inside the dapp embed.

dankins:

PR #1514, opened 2019-11-04
fix: apply to newsroom no longer checks redux

nickreynolds:
- need to  improve messaging after apply since we no longer can access data the same way, but this prevents page from crashing


nickreynolds:
yeah, the post-apply view of this page is currently broken (shows as not applied). it should be relatively easy tho to get this info from graphql, just want to get this fix up first so it's not super broken in the meantime

dankins:
:tada: This PR is included in version 4.8.33 :tada:

PR #1515, opened 2019-11-05
fix(elements): react version bump 16.11.0

dankins:
Kirby wasn't working on Firefox. The root cause was that the lazy loaded route was being loaded but not actually rendered. There must be some bug that exists only when the react root is inside an iframe.

This PR upgrades React to 16.11.0. It seems to fix it:
https://test.civil.app
I looked at the changelog and couldn't find any smoking guns. Not sure why it fixes it, but it does :man_shrugging:


dankins:
:tada: This PR is included in version 4.8.33 :tada:

PR #1516, opened 2019-11-05
refactor: payment flow

sruddy:


packages/components/src/Payments/PaymentsModal.tsx
```diff
@@ -11,10 +11,11 @@ export interface PaymentsModalProps {
 }
 
 export const PaymentsModal: React.FunctionComponent<PaymentsModalProps> = props => {
+  const maxHeight = window.innerHeight - 200;
```
tobek: Heads up/note to self that we might have to do something different here in the embed version. The embed will start out pretty small just because it doesn't need much space, and will use iframe resizer to adjust height, so in this case when payment modal is opened we'll probably want to force it to be bigger than `window.innerHeight` and resize iframe accordingly. There is a value `context.renderContext` that lets us determine if we're on the embed or dapp and update behavior based on that. I'll look into this when I'm working on iframe resizing.

sruddy: Maybe Rob can recommend a height to resize too.

dankins:

PR #1517, opened 2019-11-05
feat: first pass mobile responsive dashboard

nickreynolds:
- also fix for paramter votes showing up in 'all' when they shouldn't
- does not include mobile tabs

### DIFF
--- a/packages/dapp/src/components/Dashboard/Dashboard.tsx
+++ b/packages/dapp/src/components/Dashboard/Dashboard.tsx
@@ -4,7 +4,7 @@ import styled from "styled-components";
 import { Helmet } from "react-helmet";
 
 import { EthAddress } from "@joincivil/core";
-import { buttonSizes, Button, CivilContext, UserDashboardHeader, LoadUser } from "@joincivil/components";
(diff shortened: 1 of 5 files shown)


packages/dapp/src/components/Dashboard/Dashboard.tsx
```diff
@@ -14,9 +14,12 @@ import DashboardActivity from "./DashboardActivity";
...
 const StyledDashboardActivityContainer = styled.div`
   box-sizing: border-box;
-  padding: 0 0 200px 396px;
-  margin: -550px auto 0;
-  width: 1200px;
+  padding: 0 0 0px 0px;
+  margin-top: 40px;
+  ${mediaQueries.MOBILE} {
+    max-width: 100%;
+    margin-top: 0px;
```
sruddy: Padding can just be `padding: 0;` or you might not even need padding at all unless it's inheriting it from somewhere else. 

You can remove px in margin-top: `margin-top: 0;`.

nickreynolds:
taking care of this in more fleshed out PR

PR #1519, opened 2019-11-06
fix(dapp): don't cache index.html

dankins:
change to nginx so that index.html is sent with no-cache headers. this is definitely less than ideal since it requires the initial page load to always come from origin, however it should fix the "chunk failed to load" in (most) scenarios. 

Longer term fix is to put this in front of an edge cache, but that requires some more doing. 

Also sneaking in webpack bundle names. Pleas use this for any async loaded components going forward.


dankins:

PR #1525, opened 2019-11-07
fix(dapp): use styled-components macro for consistent class names

dankins:
In order to use automated browser testing we need to have stable class names. This requires using styled-components babel plugin, which would require ejecting from CRA. However, it looks like we can use the "macro" instead:
https://www.styled-components.com/docs/tooling#babel-macro

As far as I can tell there are no downsides to this.

Ran the following to update the files:
```
find . -type f -exec sed -i 's/import styled from "styled-components"/import styled from "styled-components\/macro"/g' {} \;
```


dankins:

PR #1527, opened 2019-11-08
A way to serve boost embed loader script [CIVIL-1243]

tobek:
This will make the standalone boost embed loader script available for newsrooms to embed from `registry.civil.co/loader/boost.js`.

It's clunky because it requires lifting information about dapp and sdk directory structure up into root package.json, but I don't see a way around that. I thought about putting this task inside dapp or sdk, bundling it with an existing build script, but this task will only work after _both_ of those packages have been built, so makes more sense for the root package.json to handle it.

I also can think of various ways to abstract multiple "loader" scripts to be served this way, but don't want to premature refactor until we see if/how we need to do something similar like this in the future, so I'm leaving it boost-specific for now.

Open to any other ways of handling this. Other ideas about it: https://civilmedia.slack.com/archives/GJ45CLYD8/p1572972104008200

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,9 @@ jobs:
                 command:  yarn build
                 no_output_timeout: 30m
             - run:
+                name: Copying boost loader script
+                command:  yarn copy:boost-loader
+            - run:
                 name: Ensuring prettified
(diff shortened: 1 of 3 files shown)


dankins:

PR #1531, opened 2019-11-12
Refactor auth modals into CivilContext and AppProvider to support login flow in story boost embed [CIVIL-1267]

tobek:
Basically we need a way to launch login/signup flow (e.g. `dispatch(await showWeb3LoginModal())`) from outside of dapp package. This commit has dapp `Web3AuthWrapper` hook these functions into `CivilContext` via `AuthService` so that `Payments` component can call it itself.

Other stuff refactored to use context instead of passing down from story feed.

End result is that story boost embed gets all this stuff for free.

### DIFF
--- a/packages/components/src/Payments/PaymentsEth.tsx
+++ b/packages/components/src/Payments/PaymentsEth.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { Mutation, MutationFunc } from "react-apollo";
 import { EthAddress } from "@joincivil/core";
+import { ICivilContext, CivilContext } from "../context";
 import { PAYMENTS_ETH_MUTATION } from "./queries";
 import { UsdEthConverter } from "../";
 import { PaymentBtn, PaymentsHide } from "./PaymentsStyledComponents";
@@ -15,11 +16,9 @@ export interface PaymentsEthProps {
   newsroomName: string;
   paymentAddress: EthAddress;
   shouldPublicize: boolean;
-  userAddress?: EthAddress;
   userEmail?: string;
   etherToSpend?: number;
   usdToSpend: number;
-  isWalletConnected: boolean;
   resetEthPayments: boolean;
   handleBoostUpdate(newUsdToSpend: number, selectedUsdToSpend: number, etherToSpend: number): void;
   handlePaymentSuccess(paymentState: PAYMENT_STATE): void;
@@ -29,9 +28,13 @@ export interface PaymentsEthStates {
   etherToSpend: number;
   usdToSpend: number;
   notEnoughEthError: boolean;
+  userAddress?: string;
 }
 
 export class PaymentsEth extends React.Component<PaymentsEthProps, PaymentsEthStates> {
+  public static contextType = CivilContext;
+  public static context: ICivilContext;
+
   public static getDerivedStateFromProps(props: PaymentsEthProps, state: PaymentsEthStates): PaymentsEthStates {
     if (props.resetEthPayments) {
       return {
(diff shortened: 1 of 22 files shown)


packages/dapp/src/components/Web3AuthWrapper.tsx
```diff
@@ -78,6 +85,20 @@ export const Web3AuthWrapper: React.FunctionComponent = () => {
     return civilContext.auth.handleInitialState();
   }
 
+  React.useEffect(() => {
+    (window as any).context = civilContext; // @TODO/tobek tmp
+    civilContext.auth.setShowWeb3Login(handleLoginClicked);
+    civilContext.auth.setShowWeb3Signup(handleSignUpClicked);
+  }, [dispatch]);
+
```
sruddy: Can you also add log out?

packages/sdk/src/react/StoryBoost/StoryBoost.tsx
```diff
@@ -109,16 +109,12 @@ export class StoryBoost extends React.Component<StoryBoostProps, StoryBoostState
                 <span>Support this newsroom</span>
                 <PaymentButton onClick={this.handleStartPayment} />
               </StoryBoostFooter>
-              <PaymentsModal open={this.state.paymentsOpen}>
+              <PaymentsModal open={this.state.paymentsOpen} handleClose={this.handleClose}>
```
sruddy: You could use `handleEndPayment` for this or `handleClose` for both.

packages/components/src/Payments/PaymentsEth.tsx
```diff
@@ -55,12 +58,22 @@ export class PaymentsEth extends React.Component<PaymentsEthProps, PaymentsEthSt
...
-    if (!this.props.isWalletConnected) {
+    const userAddress =
+      this.state.userAddress || (this.context && this.context.currentUser && this.context.currentUser.ethAddress);
+
+    if (!userAddress) {
       return (
         <PaymentsEthWrapper etherToSpend={this.state.etherToSpend} usdToSpend={this.state.usdToSpend}>
           <ConnectWalletWarningText />
-          <PaymentBtn>Select Wallet</PaymentBtn>
+          <PaymentBtn onClick={() => this.context.civil.currentProviderEnable()}>Select Wallet</PaymentBtn>
```
nickreynolds: can we update this button (in a later PR is fine) so that it becomes disabled / adds a spinner while this async call is being completed?

packages/sdk/src/react/StoryBoost/StoryBoost.tsx
tobek: Oh right, whoops, I didn't look at how `handleEndPayment` was being used and thought it'd be safer to keep "close" separate from "end payment" in case one had to change, but uh yeah now I see.

packages/dapp/src/components/Web3AuthWrapper.tsx
tobek: Good call, just went looking for it and found that it's actually already there at `context.auth.logout()`, set by `AuthService`.

packages/components/src/Payments/PaymentsEth.tsx
tobek: If a user has previously connected wallet and is logged in to wallet but for some reason eth isn't enabled (happens) then this call is basically instantaneous. If MM (haven't tested portis) isn't connected and/or not unlocked then it pops up MM. If the user closed or didn't see MM popup and we put a disabled loading state on button, they'd be kind of stuck. As it is now, they could at least click button again and maybe get popup again? Not sure what the best UX is here.

dankins:
:tada: This PR is included in version 4.9.2 :tada:

PR #1541, opened 2019-11-16
Share network/gql logic between dapp and embeds [CIVIL-1283]

tobek:
Moving some stuff from `Main` (dapp-only) to `Web3AuthWrapper` and `AppProvider` (both dapp and embeds). This fixes staging-only bug where if the embed is the first time you visit the dapp then the network value is set wrong, also fixes issues where iframes have isolated localStorage as Safari does.

Also updating logic so that gql endpoint depends on `config.DEFAULT_ETHEREUM_NETWORK` to begin with, but it won't overwrite network value set by web3 network being updated. (Though we still don't, I think, change the uri in already-created apollo client, but on page refresh the network value will take effect).


dankins:

PR #1547, opened 2019-11-20
feat: show user profile on dashboard with ability to edit

nickreynolds:
- editing/setting email and avatar from dashboard is functional, but not implemented as designed (since design is not ready yet)
- display of email address is not as designed (since designs not updated to reflect this field yet)

### DIFF
--- a/packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
+++ b/packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
@@ -78,7 +78,95 @@ export const StyledDashboardActivityDescription = styled.div<StyledDashboardActi
 `;
 
 export const StyledUserInfo = styled.div`
-  width: 277px;
+  width: 333px;
+  ${mediaQueries.MOBILE} {
+    width: 100%;
+  }
+`;
+
+export const StyledUserProfile = styled.div`
+  width: 333px;
+  ${mediaQueries.MOBILE} {
+    width: 100%;
+  }
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  margin-bottom: 22px;
+`;
+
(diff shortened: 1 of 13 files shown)


packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
```diff
@@ -78,7 +78,95 @@ export const StyledDashboardActivityDescription = styled.div<StyledDashboardActi
 `;
 
 export const StyledUserInfo = styled.div`
-  width: 277px;
+  width: 333px;
+  ${mediaQueries.MOBILE} {
```
sruddy: FYI `mediaQueries.MOBILE` is `max-width: 1030px`, `mediaQueries.MOBILE_SMALL` is `max-width: 640px` not sure you ant the larger size here. We should name these better when we make updates to style guide.

packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
```diff
@@ -78,7 +78,95 @@ export const StyledDashboardActivityDescription = styled.div<StyledDashboardActi
...
+    width: 100%;
+  }
+`;
+
+export const StyledUserProfile = styled.div`
+  width: 333px;
+  ${mediaQueries.MOBILE} {
```
sruddy: Same here

packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
nickreynolds: `MOBILE` is the point where we change the layout of the page (stack components on top  of each other), so I think it's correct here

dankins:

PR #1548, opened 2019-11-20
feat: tasks dropdown

nickreynolds:
- replace "tasks" subtabs with a dropdown + button (for token transfer)
- modifications to how width is controlled so all dashboard tabs contents display with same width
- the changes to MyTasks are a little bit hacky, but it's a very unique one-off component, and I'm not sure it's worth spending time making it very reusable

### DIFF
--- a/packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
+++ b/packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
@@ -264,6 +264,7 @@ export const StyledTransferTokenTitle = styled.div`
   text-align: center;
   margin: 0 0 40px;
   padding: 20px;
+  flex-grow: 1;
 
   h3 {
     font-size: 20px;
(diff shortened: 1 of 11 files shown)


packages/elements/src/colors/index.ts
```diff
@@ -29,6 +29,7 @@ export const colors = {
     CIVIL_GRAY_4: "#E9E9EA",
     CIVIL_GRAY_5: "#F4F4F4",
     CIVIL_GRAY_6: "#FCFCFC",
+    CIVIL_GRAY_7: "#9B9B9B", // NewPrimaryButton border color
```
tobek: Aw man now the grays aren't numbered in increasing lightness, but I guess no way around it unless `CIVIL_GRAY_2A` or something, which would be dumb.

packages/components/src/TCRUserDashboard/DashboardStyledComponents.tsx
```diff
@@ -352,6 +352,7 @@ export const StyledTransferTokenTitle = styled.div`
   text-align: center;
   margin: 0 0 40px;
   padding: 20px;
+  flex-grow: 1;
```
tobek: Just to confirm, are these guys constrained to any max-width anywhere or by their parent? Because on very wide screen full width looks a bit wack.

nickreynolds: I'll add a max-width to the whole thing eventually, but we already had some things that weren't constrained at all and it was way worse when they were different sizes. At least now they're all the same size

dankins:

PR #1559, opened 2019-11-25
fix: listing history tweaks

nickreynolds:


packages/components/src/ChallengeResultsChart/styledComponents.tsx
```diff
@@ -72,7 +74,7 @@ export const VotesPerTokenVote = styled.div<VotesPerTokenVoteProps>`
         ? colors.accent.CIVIL_TEAL
         : colors.accent.CIVIL_RED};
     margin-right: 4px;
-  ]
+  }
```
tobek: :exclamation: good catch. It's a bummer that we don't (can't?) have linting inside styled components. I've seen pretty often for instance a missing semicolon that silently borks the rules before and after.

dankins:

PR #1563, opened 2019-12-03
Slightly better post-apply flow

tobek:
Had this lying around on my computer, not sure if it really matters since I think no one will go all the way through this flow anymore, but it's a trivial update that is a bit better.

> With redux stuff removed, there's no feedback at all after applying to registry, it looks like you never did it. This way we can at least direct the user to the dashboard where they can see they've applied.


dankins:

PR #1569, opened 2019-12-05
fix(dapp): don't require injected ethereum

dankins:
this is needed so that Portis works


dankins:
> this is after the user has already selected a wallet, right?

Yes, pretty sure that happens on the prior screen.

dankins:

PR #1582, opened 2019-12-16
feat: updates to apply flow to reflect new fastpass process

nickreynolds:


packages/newsroom-signup/src/NewsroomProfile/GrantApplication.tsx
```diff
@@ -224,35 +230,17 @@ class GrantApplicationComponent extends React.Component<GrantApplicationProps &
...
-            return (
-              <ButtonSection>
-                <BorderlessButton onClick={this.deselectSkip}>Cancel</BorderlessButton>
-                <BorderlessButton
-                  onClick={async () => {
-                    this.props.dispatch!(grantSkipped());
-                    return requestGrant({
-                      variables: {
-                        input: false,
-                      },
```
tobek: At some point we had a bug that erroneously let users pass the grant step (by clicking on the section nav) and so we added logic higher up in component tree that would force the user back to the grant step if a) they are past the grant step, and b) `grantRequested` isn't explicitly set to false. With this mutation removed I imagine this will happen. Have you tested skipping the grant step and moving forward in the process but then reloading the page? Might have to either add this mutation back in or disable that logic, though that logic ended up being kind of complex IIRC.

nickreynolds: just  tested, if you skip the grant and reload the page, it will take you back to the "choose between grant and skip" if you're currently on a step BEFORE creating your Newsroom smart contract. Once you've created that contract, reloading will take you to the "tutorial" or beyond. This seems basically like the correct behavior

But yes some of this logic is pretty wild so we should test a few more times once it's on staging

dankins:
:tada: This PR is included in version 4.9.5 :tada:

PR #1593, opened 2020-01-04
Get Started page [CIVIL-1289]

tobek:


packages/dapp/src/components/GetStarted.tsx
```diff
@@ -0,0 +1,248 @@
...
+  font-size: 13px;
+`;
+
+const GetStartedPage = () => {
+  const context = React.useContext<ICivilContext>(CivilContext);
+  if (!context || !context.auth) {
+    return <LoadingMessage />;
+  }
+  if (context.currentUser) {
+    // @TODO/tobek For logged-in users there's a flash of the page before redirect; there shouldn't be.
```
nickreynolds: @dankins this is probably because the user is being loaded, right? so for a tiny bit there's no `currentUser`, then the graphql stuff finishes and sets the user in context, right? is there a currentUserLoading?

tobek: That is exactly it. However, 1) there is a `context.auth.loading` flag I assume is for this use case so I'll add a check for that in, and 2) even without using that, I'm not seeing this flash any more so... that's cool.

dankins:
:tada: This PR is included in version 4.9.5 :tada:

PR #1613, opened 2020-01-16
feat: newsroom handles

nickreynolds:
- allow listing pages to load data with listing address or handle in URL
- link to listing page with handle from whitelisted cards, boosts
- navigate to appropriate page with address instead of handle when user tries to challenge listing loaded via handle


dankins:
:tada: This PR is included in version 4.9.7 :tada:

PR #1614, opened 2020-01-17
fix: remove fields causing errors on storyfeed boost query

nickreynolds:


packages/dapp/src/components/StoryFeed/queries.ts
```diff
@@ -37,7 +34,7 @@ export const STORY_FEED_QUERY = gql`
           }
           ... on PostExternalLink {
             id
-            datePosted
+            publishedTime
```
sruddy: what's the difference between `publishedTime` and `createdAt`?

nickreynolds: so publishedTime is a new field that represents EITHER the opengraph article `published_time` or the value from the civil published_time meta tag that the WP plugin now adds to posts. the storyfeed should be updated to use this value instead of the opengraph value

it can still be null if the article has neither OG published time or Civil published time

dankins:

PR #1615, opened 2020-01-17
feat(kirby): use same signature for login and signup

dankins:
This changes the signature we present to the user to log in sign up. Previously was "Sign up to Civil @ {timestamp}" or "Log in to Civil @ {timestamp}". Now it is "Authenticate to {domain} @ {timestamp}". This allows us to reuse the signature if an account already exists (signup -> login). Additionally, it makes it clearer when authenticating to 3rd party sites-  you aren't signing up or logging in (from Kirby's perspective), you are just authenticating and the provider decides what to do with the signature. 

NOTE: this needs to ride with civil-api-server changes https://github.com/joincivil/civil-api-server/pull/324

### DIFF
--- a/packages/kirby/src/plugins/CivilID.ts
+++ b/packages/kirby/src/plugins/CivilID.ts
@@ -11,12 +11,12 @@ export interface ChildPluginState {
   pendingLoginRequest?: {
     requestID: number;
     message: string;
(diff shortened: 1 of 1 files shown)


dankins:

PR #1619, opened 2020-01-22
fix: better permissions checks and UI for Manage Newsroom [CIVIL-918]

tobek:
Whole Manage Newsroom page should be blocked if not `currentUserIsAdmin`, etc.


tobek:
Boost Proceeds are shown on the newsroom dashboard, which already is set up to only show newsrooms you're an admin of, so no way to check this without graphql playground or some temp code change to dashboard. This update is to block off the `/manage-newsroom` route to non-admins, which has edit charter, add smart contract roles, and launch project boost.

dankins:

PR #1621, opened 2020-01-22
Add Sentry to boost loader [CIVIL-1434] + misc Story Boost embed fixes

tobek:
The Story Boost embed itself loads the whole dapp wrapper in an iframe and thus loads Sentry. However, the (relatively) minimal JS code that we actually load on a newsroom's page that creates the ExternalLink post and loads the iframe, that did not have Sentry.

Errors in the embed JS are important and need logging.

Adding Sentry brings the embed JS from 409kb (115kb gzipped) to 469kb (131kb gzipped) so it's not great but not terrible.

Having to add `SENTRY_DSN` constant manually to `sdk` package isn't great, but not sure how else to do this. Thought about adding it to `utils` package but the embed loader doesn't load `utils` and doing so would of course bring some junk in. There's no server-side environment that can add this in, it's a static built JS file that we serve via the dapp server.

### DIFF
--- a/packages/dapp/src/helpers/config.ts
+++ b/packages/dapp/src/helpers/config.ts
@@ -13,7 +13,6 @@ export interface ConfigType {
   SENTRY_DSN?: string;
   STRIPE_CLIENT_ID?: string;
   STRIPE_API_KEY?: string;
-  APP_VERSION?: string;
   KIRBY_TARGET_ORIGIN?: string;
   KIRBY_IFRAME_SRC?: string;
 }
@@ -33,7 +32,6 @@ const defaultConfig: ConfigType = {
   SENTRY_DSN: process.env.REACT_APP_SENTRY_DSN,
   STRIPE_CLIENT_ID: process.env.REACT_APP_STRIPE_CLIENT_ID,
   STRIPE_API_KEY: process.env.REACT_APP_STRIPE_API_KEY,
-  APP_VERSION: process.env.REACT_APP_APP_VERSION,
   KIRBY_TARGET_ORIGIN: process.env.REACT_APP_KIRBY_TARGET_ORIGIN,
   KIRBY_IFRAME_SRC: process.env.REACT_APP_KIRBY_IFRAME_SRC,
 };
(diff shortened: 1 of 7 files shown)


packages/dapp/src/helpers/config.ts
```diff
@@ -13,7 +13,6 @@ export interface ConfigType {
   SENTRY_DSN?: string;
   STRIPE_CLIENT_ID?: string;
   STRIPE_API_KEY?: string;
-  APP_VERSION?: string;
```
tobek: This is not supplied by env variables in neither staging nor prod. I grepped repos for whole joincivil organization and didn't see `APP_VERSION` used anywhere.

packages/sdk/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const SENTRY_DSN = "https://4659311eb5cb472982a5f7d392f3e2b4@sentry.io/1430767";
```
dankins: I agree that it is definitely less than ideal having this there, but no easy solve right now. imo, the sdk package is not really necessary now that we are serving everything from an iframe and not directly on the site. I think the solve here is to just move all `react` stuff in sdk back in to the dapp. 

Make just throw a TODO in explaining some of the thought process?

packages/dapp/src/helpers/config.ts
tobek: It's currently referenced only one place, in the Sentry init config, where it's currently undefined, which is identical to not including the key in the first place (unless Sentry is doing something very weird iterating over keys).

But I agree this could be useful if the container supplied a value for it, I'll leave it in for now.

packages/sdk/src/constants.ts
tobek: The SDK package is still used to build this (relatively) minimal JS script that does run directly on the newsroom site in order to run the iframe resizer script (also does some other stuff - the stuff that can generate errors Sentry needs to collect - that in retrospect should be refactored and moved into the iframe).

I'll add a TODO and also make a ticket for refactoring some stuff so that the logic that needs Sentry happens inside the iframe.

dankins:

PR #1623, opened 2020-01-24
Story Boost embeds: AvatarLogin, header, and footer updates [CIVIL-1370]

tobek:


packages/components/src/Payments/PaymentsWrapper.tsx
```diff
@@ -67,7 +68,9 @@ export class PaymentsWrapper extends React.Component<PaymentsWrapperProps> {
       <>
         {this.props.boostType !== "project" && <SendPaymentHdrText />}
         <PaymentHeaderFlex>
-          <PaymentHeaderNewsroom>{this.props.newsroomName}</PaymentHeaderNewsroom>
+          <PaymentHeaderNewsroom>
+            <StoryNewsroomStatus newsroomName={this.props.newsroomName} activeChallenge={false} />
```
sruddy: 🤦‍♀ I think this is my mistake `activeChallenge={false}` is fake data need to pull in challenge data from newsroom. I think this is likely an issue on all `StoryNewsroomStatus` I can create a bug ticket

tobek: Aha no worries yeah that makes sense, I actually just assumed it was a placeholder and that we hadn't yet implemented various bits of logic needed here. Not sure if whitelisted flag for a particular newsroom will be in scope in the various places we need it to be so it might be a bit of a refactor (possibly we could pass newsroom address into `StoryNewsroomStatus` and have the component itself fetch the status).

dankins:

PR #1646, opened 2020-02-20
Save Cards

nickreynolds:
- create a new stripe form that handles paying via payment intents
- allows a user to save a new card to their (civil) customer, or choose a card saved on their (civil) customer, or enter new card without saving it during payment flow. Then this payment method is cloned to the connected (newsroom) stripe account, a payment intent is created and confirmed on the newsroom stripe account
- does not include adding/removing/setting-as-default from the accounts page's payment methods tab


dankins:
:tada: This PR is included in version 4.10.2 :tada:

PR #1651, opened 2020-02-28
fix(sdk): don't load boost embed if it's not in <body>

tobek:
Prevent out of body boost experience as happening on Gotham Gazette. This check happens in embed loader, outside of embed iframe, and so won't blow up sentry.


dankins:
:tada: This PR is included in version 4.10.3 :tada:

